PR #5, opened 2017-03-04
added draw polygon function

tejus-gupta:
Function to draw polygon. It takes an array of CartesianIndex as input.

### DIFF
--- a/src/paths.jl
+++ b/src/paths.jl
@@ -10,9 +10,9 @@ closed (keyword): whether to connect first and last points
 In case a point is out-of-bounds, it throws an error after drawing the line segments till that point.
 """
 
-lines{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T=one(T); closed::Bool=true) = lines!(copy(img),vertices,color,closed=closed)
+path{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T=one(T); closed::Bool=true) = path!(copy(img),vertices,color,closed=closed)
 
-function lines!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T=one(T); closed::Bool=true)
+function path!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T=one(T); closed::Bool=true)
     f = CartesianIndex(map(r->first(r)-1, indices(img)))
     l = CartesianIndex(map(r->last(r), indices(img)))
 
--- a/src/polygon2d.jl
+++ b/src/polygon2d.jl
@@ -1,0 +1,43 @@
+"""
+```
+draw_polygon_perimeter!(img, vertices)
+img2=draw_polygon_perimeter(img, vertices)
+```
+Draws the perimeter of the polygon formed by input vertices(Array of CartesianIndex).
+"""
+
+function draw_polygon_perimeter{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1})
+
+    f = CartesianIndex(map(r->first(r)-1, indices(img)))
+    l = CartesianIndex(map(r->last(r), indices(img)))
+    for vertex in vertices
+        if min(f, vertex)!=f || max(l, vertex)!=l
+            println(vertex)
+            error("Polygon coordinates out of range.")
+        end
+    end
+
+    img2 = zeros(T, size(img))
+    for i in 1:length(vertices)-1
+        line!(img2, vertices[i], vertices[i+1])
+    end
+    line!(img2, vertices[1], vertices[end])
+    img2
+end
+
+function draw_polygon_perimeter!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1})
+
+    f = CartesianIndex(map(r->first(r)-1, indices(img)))
+    l = CartesianIndex(map(r->last(r), indices(img)))
+    for vertex in vertices
+        if min(f, vertex)!=f || max(l, vertex)!=l
+            println(vertex)
+            error("Polygon coordinates out of range.")
+        end
+    end
+
+    for i in 1:length(vertices)-1
+        line!(img, vertices[i], vertices[i+1])
+    end
+    line!(img, vertices[1], vertices[end])
+end
--- a/src/polygon2d.jl
+++ b/src/polygon2d.jl
@@ -1,16 +1,18 @@
 """
 ```
 draw_polygon_perimeter!(img, vertices)
+draw_polygon_perimeter!(img, vertices, color)
 img2=draw_polygon_perimeter(img, vertices)
+img2=draw_polygon_perimeter(img, vertices, color)
 ```
 Draws the perimeter of the polygon formed by input vertices(Array of CartesianIndex).
 """
 
-
-
 draw_polygon_perimeter{T<:Colorant}(img::AbstractArray{T, 2}, args...) = draw_polygon_perimeter!(copy(img), args...)
 
-function draw_polygon_perimeter!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1})
+draw_polygon_perimeter!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1})=draw_polygon_perimeter!(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, one(T))
+
+function draw_polygon_perimeter!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T)
 
     f = CartesianIndex(map(r->first(r)-1, indices(img)))
     l = CartesianIndex(map(r->last(r), indices(img)))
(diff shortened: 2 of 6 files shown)


tejus-gupta:
Should I change the return type to the perimeter vertices stored as a CartesianIndex array. The user can then mark those points on the images.

src/polygon2d.jl
```diff
@@ -0,0 +1,43 @@
+"""
+```
+draw_polygon_perimeter!(img, vertices)
+img2=draw_polygon_perimeter(img, vertices)
+```
+Draws the perimeter of the polygon formed by input vertices(Array of CartesianIndex).
+"""
+
+function draw_polygon_perimeter{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1})
```
mronian: You could use the `draw_polygon_perimeter!` function by passing `copy(img)` as the first argument to avoid duplicating code. You could check out how the `line` function is implemented.

test/polygon2d.jl
```diff
@@ -0,0 +1,23 @@
...
+    @test all(x->x==true, img[1:3,1])==true
+    @test all(x->x==true, img[3,1:3])==true
+    @test img[2,4]==true
+
+    img=zeros(ColorTypes.Gray{Bool},5,5)
+    draw_polygon_perimeter!(img, vert)
+    @test all(x->x==true, img[1,:])==true
+    @test all(x->x==true, img[1:3,1])==true
+    @test all(x->x==true, img[3,1:3])==true
+    @test img[2,4]==true
```
mronian: Needs some tests to check that the inner area is empty.

mronian:
@tejus-gupta Looks good 😄 Would you like to work on adding a `thickness` parameter to all the functions. A negative value for thickness will fill the figure from inside and a positive value from outside.

test/polygon2d.jl
tejus-gupta: I have added some test for that.

tejus-gupta:
Yes, that would be a nice function. But I wanted to add a draw perimeter and fill polygon function first.

src/polygon2d.jl
```diff
@@ -0,0 +1,28 @@
...
+draw_polygon_perimeter!(img, vertices)
+img2=draw_polygon_perimeter(img, vertices)
+```
+Draws the perimeter of the polygon formed by input vertices(Array of CartesianIndex).
+"""
+
+
+
+draw_polygon_perimeter{T<:Colorant}(img::AbstractArray{T, 2}, args...) = draw_polygon_perimeter!(copy(img), args...)
+
```
mronian: Lets add a `color` parameter here. `line` supports it.

tejus-gupta: I have added the color parameter.

test/polygon2d.jl
```diff
@@ -0,0 +1,25 @@
...
+    @test all(x->x==false, img[2,2:3])==true
+    @test img[2,4]==true
+
+    img=zeros(ColorTypes.Gray{Bool},5,5)
+    draw_polygon_perimeter!(img, vert)
+    @test all(x->x==true, img[1,:])==true
+    @test all(x->x==true, img[1:3,1])==true
+    @test all(x->x==true, img[3,1:3])==true
+    @test all(x->x==false, img[2,2:3])==true
+    @test img[2,4]==true
```
mronian: Could you add one test for color?

src/polygon2d.jl
```diff
@@ -0,0 +1,30 @@
...
+function draw_polygon_perimeter!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T)
+
+    f = CartesianIndex(map(r->first(r)-1, indices(img)))
+    l = CartesianIndex(map(r->last(r), indices(img)))
+    for vertex in vertices
+        if min(f, vertex)!=f || max(l, vertex)!=l
+            println(vertex)
+            error("Polygon coordinates out of range.")
+        end
+    end
```
mronian: @timholy Should we assume that the input will be valid? I am concerned about the speed of this package since it might be used with streaming video. 

@tejus-gupta could you post some times with and without these checks? Also, see if adding checks to the `line` function is making it too slow.

src/polygon2d.jl
```diff
@@ -0,0 +1,30 @@
+"""
+```
+draw_polygon_perimeter!(img, vertices)
```
timholy: Since `draw` is in the name of the package, let's leave it out of the name of the function.

src/polygon2d.jl
timholy: Currently `line!` effectively checks for this because `img[x0, y0] = color` is not inside of an `@inbounds`. However, compared to checking each pixel, checking the endpoints is much more efficient. So it would be better to check this way and add `@inbounds` to the loop of `bresenham`. I think that would also work for `xiaolin_wu` but I'd want someone to take a closer look to be sure that roundoff error wouldn't cause problems.

timholy: One thing worth asking, do we want it to skip drawing the polygon if one of the vertices is out of range? Or should we draw up to the point of the error? If the latter, then the check could be placed inside `line!` or the individual routines.

tejus-gupta:
With checks on vertices-
with0

Without checks-
with1

The run time was almost identical on average.

test/polygon2d.jl
tejus-gupta: Done :)

timholy:
Nice. For such short execution times, I recommend [BenchmarkTools](https://github.com/JuliaCI/BenchmarkTools.jl). Make sure you use interpolation, i.e.,
```julia
@benchmark polygon_perimeter!($img, $vert3)
```
That said, this looks fine as is (more for future reference).

src/polygon2d.jl
mronian: I think we should draw upto the point of error. @tejus-gupta what do you say?

tejus-gupta: We can do either of these two things-
1. Draw all edges except those containing outside vertices.
option2
2. Only draw part of polygon inside the image.(white rectangle is image)
option3

I think we should give a warning to the user in either case.

timholy: I don't like warnings; suppose the user deliberately wanted your second option, it would be really annoying to have a warning pop up each time. I think if folks want that version, we should support a "clipping path" that breaks the path up into in-bounds segments that terminate when they get to the edge:
```julia
segments = clip(vertices, indices(img))
for seg in segments
    polygon_perimiter!(img, seg)
end
```
Otherwise, I think we should simply throw an error if there's an out-of-bounds segment. We could either do that
- before drawing anything at all
- as soon as the first out-of-bounds vertex is encountered (which terminates future edges even if they are in bounds)
- your option 1

Of these, I like the first and third better than the middle one. The middle one would be the absolute highest performance option, but compared to the cost of actually drawing the pixels I doubt we would notice.

tejus-gupta: The clipping wouldn't work because polygon_perimeter! will also connect the first and last vertex in each segment.

timholy: Good point. `lines!(img, seg)`?

tejus-gupta: I also like the first(current version) and third option better.

timholy: I wonder if we should call this `lines!(img, seg, closed=true)`?

tejus-gupta: That looks better. What do you think @mronian? Should I update the pull request? We still need to decide when to throw the out-of-bounds error.

mronian: Yeah. `lines!(img, seg, closed=true)` looks good to me.

tejus-gupta: I have updated the function to lines(img, vert, color, closed=true)
In case a point is out-of-bounds, it throws an error after drawing the line segments till that point.

tejus-gupta:
I have updated the function to lines(img, vert, color, closed=true).
In case a point is out-of-bounds, it throws an error after drawing the line segments till that point.

mronian:
Should we move this to `line2d.jl`?

tejus-gupta:
I think that functions that use line! to create simple utilities like rectangles, arrowed line, polygon perimeter, fill polygon should be in a separate file. Any suggestions for better name than polygon2d.jl?

timholy:
Maybe `paths.jl`? In which case maybe we should rename it to `path!(img, vertices, closed=true)`?

tejus-gupta:
I have changed the function to path(img, vertices, color; closed=true) and the file name to paths.jl.

src/paths.jl
```diff
@@ -0,0 +1,36 @@
...
+vertices: coordinates of points as array of CartesianIndex
+color (optional): color of line segments
+closed (keyword): whether to connect first and last points
+
+In case a point is out-of-bounds, it throws an error after drawing the line segments till that point.
+"""
+
+path{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T=one(T); closed::Bool=true) = path!(copy(img),vertices,color,closed=closed)
+
+function path!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianIndex{2},1}, color::T=one(T); closed::Bool=true)
```
Evizero: Would it be worth considering making `closed` a positional argument? I did not benchmark, but I suspect that drawing a lot of paths in a loop could mean that invoking this function is sensitive to the keyword-argument overhead

Evizero: one could then create a convenience method `path(img, vertices; color=.., closed=...)` which in turn calls method with positional arguments.

Evizero: That said I don't think that we need to make that decision right now. So don't consider my suggestion as merge-blocking

timholy: Those are good thoughts. I'd say, let's use the API we want and then change later if we discover the performance isn't acceptable. By the time someone notices a problem, one can always hope that the penalty for keyword arguments will be gone anyway.

mronian:
Looks good for a merge. I'll wait for sometime in case anyone would like to suggest something and then merge it.

mronian:
Thanks @tejus-gupta 😄

PR #9, opened 2017-03-21
Added line_normal method

annimesh2809:
Draws a line on image, given tuple τ of the form of (ρ,θ) for the normal form of line:
    x*cos(θ) + y*sin(θ) = ρ

Works nicely with `hough_transform_standard`:
Original image:
org

After applying `canny`:
canny

Applied `hough_transform_standard` and then using `line_normal`:
ht

### DIFF
--- a/src/line2d.jl
+++ b/src/line2d.jl
@@ -104,3 +104,50 @@ function xiaolin_wu{T<:Gray}(img::AbstractArray{T, 2}, y0::Int, x0::Int, y1::Int
     end
     img
 end
+
+"""
+```
+img_with_line = line_normal(img, τ, color, method)
+img_with_line = line_normal(img, ρ, θ, color, method)
+```
+
+Draws a line on the input image given tuple `τ` of the form (ρ,θ) for the normal form of line:
+    x*cos(θ) + y*sin(θ) = ρ
+with the given `color`. Lines are drawn using the `bresenham` method by default. If anti-aliasing
+is required, the `xiaolin_wu` can be used.
+"""
+line_normal{T<:Colorant}(img::AbstractArray{T, 2}, args...) = line_normal!(copy(img), args...)
+
+line_normal!{T<:Colorant}(img::AbstractArray{T, 2}, τ::Tuple{Number,Number}, method::Function = bresenham) = 
+    line_normal!(img, τ[1], τ[2], one(T), method)
+
+line_normal!{T<:Colorant}(img::AbstractArray{T, 2}, τ::Tuple{Number,Number}, color::T, method::Function = bresenham) = 
+    line_normal!(img, τ[1], τ[2], color, method)
+
+line_normal!{T<:Colorant}(img::AbstractArray{T, 2}, ρ::Number, θ::Number, method::Function = bresenham) =
+    line_normal!(img, ρ, θ, one(T), method)
+
+function line_normal!{T<:Colorant}(img::AbstractArray{T, 2}, ρ::Number, θ::Number, color::T, method::Function = bresenham)
+    h,w = size(img)
+    ρ -= 1; h -= 1; w -= 1
+    t = Vector{Number}(4)
+    cosθ = cos(θ)
+    sinθ = sin(θ)
+    t[1] = ( ρ * cosθ    )/ sinθ
+    t[2] = (-ρ * sinθ    )/ cosθ
+    t[3] = ( ρ * cosθ - w)/ sinθ
+    t[4] = (-ρ * sinθ + h)/ cosθ
+    hasInf = false
+    for i in 1:4
+        if ! isfinite(t[i])
+            if ! hasInf
+                t[i] = Inf
+                hasInf = true
+            else
+                t[i] = -Inf
+            end
+        end
+    end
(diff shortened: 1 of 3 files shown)


mronian:
Why not call it `line` also?

src/line2d.jl
```diff
@@ -104,3 +104,50 @@ function xiaolin_wu{T<:Gray}(img::AbstractArray{T, 2}, y0::Int, x0::Int, y1::Int
...
+    line_normal!(img, τ[1], τ[2], one(T), method)
+
+line_normal!{T<:Colorant}(img::AbstractArray{T, 2}, τ::Tuple{Number,Number}, color::T, method::Function = bresenham) = 
+    line_normal!(img, τ[1], τ[2], color, method)
+
+line_normal!{T<:Colorant}(img::AbstractArray{T, 2}, ρ::Number, θ::Number, method::Function = bresenham) =
+    line_normal!(img, ρ, θ, one(T), method)
+
+function line_normal!{T<:Colorant}(img::AbstractArray{T, 2}, ρ::Number, θ::Number, color::T, method::Function = bresenham)
+    h,w = size(img)
```
timholy: One of my next projects is to provide better support for visualizations with [arrays that are indexed at arbitrary values](http://docs.julialang.org/en/latest/devdocs/offset-arrays.html#Arrays-with-custom-indices-1); this is quite useful for, e.g., image transformations (rotation, warping, etc), where having arbitrary indices [is helpful for keeping track of actual location](http://juliaimages.github.io/latest/indexing.html).

Can this be changed to `indsy, indsx = indices(img)`?

src/line2d.jl
```diff
@@ -104,3 +104,50 @@ function xiaolin_wu{T<:Gray}(img::AbstractArray{T, 2}, y0::Int, x0::Int, y1::Int
...
+    ρ -= 1; h -= 1; w -= 1
+    t = Vector{Number}(4)
+    cosθ = cos(θ)
+    sinθ = sin(θ)
+    t[1] = ( ρ * cosθ    )/ sinθ
```
timholy: I wonder if it would be clearer to compute the 4 intersection points between the (extended) line and the (extended) lines that form the boundaries of the image:
```julia
intersections_x = [(x, (ρ - x*cosθ)/sinθ) for x in (first(indsx), last(indsx))] # with the left and right edge lines
intersections_y = [((ρ - y*sinθ)/cosθ, y) for y in (first(indsy), last(indsy))] # with the top and bottom edge lines
```
and then keeping the 0 or 2 intersections that occur within the bounds of the image. (What happens with the current implementation if `ρ, θ` are such that the line doesn't intersect the image?)

If you like this parametrization, note there are no subtractions of 1 like on line 132.

test/line2d.jl
```diff
@@ -150,4 +150,35 @@
...
+		end
+		@test line_normal(img, 7.5, π/4) == expected
+		for i in 1:10
+			expected[i,10-i+1] = 0.8
+		end
+		fill!(expected, Gray(0))
+		for i in 1:10
+			expected[i,i] = Gray(0.8)
+		end
+		@test line_normal(img,(1,-π/4), Gray(0.8)) == expected
```
timholy: Would be good to add a test for the case when the parameters don't intersect the image.

src/line2d.jl
```diff
@@ -104,3 +104,50 @@ function xiaolin_wu{T<:Gray}(img::AbstractArray{T, 2}, y0::Int, x0::Int, y1::Int
...
+            if ! hasInf
+                t[i] = Inf
+                hasInf = true
+            else
+                t[i] = -Inf
+            end
+        end
+    end
+    sort!(t)
+    method(img, round(Integer, ρ*sinθ + t[2]*cosθ + 1), round(Integer, ρ*cosθ - t[2]*sinθ + 1), round(Integer, ρ*sinθ + t[3]*cosθ + 1), round(Integer, ρ*cosθ - t[3]*sinθ + 1), color)
```
timholy: If there's anti-aliasing, do we want to round the vertices? It seems that passing a fractional value would increase the accuracy.

annimesh2809: Yes, that is correct. But the anti-aliasing line method `xiaolin_wu` takes `Int` coordinates of points. @timholy Should we modify the function `xiaolin_wu` to take `Number` rather than `Int` and do a similar change in `bresenham` for consistency?

annimesh2809:
@mronian As @timholy mentioned that we should change the name to `line` only if we are intending to create more parametrized version of `line`. But I can't think of any other parameterized form of `line` that would be necessary. Should I rename the method to `line` in case some other parametrizations might be needed in future or leave it to `line_normal`?

kmsquire:
Just a vote for renaming to `line`. No strong reason, but that seems more Julian to me.

src/line2d.jl
timholy: I don't have strong feelings about it, but it seems somewhat better to me.

annimesh2809:
@timholy I have changed the method name to `line!` and added some wrapper types for input to these methods.

timholy:
Thanks, this is great!

PR #10, opened 2017-04-01
Fix CI testing

timholy:


Evizero:
Is `Images` really needed? It looks more like a rouge `using` to me https://github.com/JuliaImages/ImageDraw.jl/blob/master/test/paths.jl#L1

Evizero:
Just tested it locally. removing `Images` works

Evizero:
Ah, you beat me to it :)

timholy:
You're so right. I think it was being used to load `ColorTypes` and `FixedPointNumbers` if that file was run as a standalone.

Issue #11, opened 2017-04-05
A better structure for `ImageDraw.jl`


annimesh2809:
Currently, there are some structural and naming issues in `ImageDraw.jl`:
1. `line` is a method for drawing a line while `Line` is an abstract type denoting a line.
2. `line` used with `LineTwoPoints` draws only a line segment while `line` used with `LineNormal` draws an infinite line on the image, which breaks consistency.
3. Function names like `line`, `ellipse`, `path` etc. do not clearly suggest whether they are methods to draw an object on an image or methods to create a new drawable object.

I suggest the following changes in the API:
1. Any object that can be drawn on an image like `Line`, `Ellipse`, `Polygon` etc. should derive from `Drawable` abstract type.
2. A single method `draw` to draw `Drawable` object(s) on an image(or multiple images). `draw` would use multiple dispatch to draw the specific objects onto the image. This would produce a much neater API.
3. The addition of `RegularPolygon <: Drawable` to draw a regular polygon on an image. This could be used to define other shapes like `Square`, `EquilateralTriangle` etc.
4. `Line` would be used to draw infinite lines and `LineSegment` would be used to draw a line segment of a specific length.
![imagedraw jl](https://cloud.githubusercontent.com/assets/15063205/24719150/5234c79e-1a56-11e7-9828-0c390e28ddef.png)

Kindly suggest if these changes are relevant. If so, then I can start working on this.

timholy:
I like it! Including the diagram.

My only tiny suggestion would be to call the method `draw!`, since it modifies the image. Of course you could provide a copying `draw` function built on top of `draw!`. But the non-copying version will be much more useful if you plan to do a lot of drawing on a single image.

mronian:
Nice idea @annimesh2809!

Evizero:
I am also a fan. I am looking forward to this

annimesh2809:
Thanks for encouraging my idea. I have started with the implementation.
As `ImageDraw.jl` is not a registered package, should we add deprecation warnings for usage of old syntax?

timholy:
If it's not hard, by all means add them, but since it's not registered I don't think it's required.

bjarthur:
should this be closed now that #13 is merged?

timholy:
Yeah. We should register this package now, I think. I guess it needs docs, though.

annimesh2809:
[...]

PR #13, opened 2017-04-13
Modified structure of ImageDraw.jl

annimesh2809:
As mentioned in #11:
* All drawable objects are derived from `Drawable` abstract type.
* `draw!` (or `draw`) can be used to draw one or more `Drawable` objects on an image.
```
center = Point(200,200)
img = zeros(RGB, 400, 400)
imshow(draw(img, [CirclePointRadius(center, 50), center, LineNormal(150,0)], [RGB(1,0,0), RGB(0,0,1), RGB(1,0,1)]))
```
img1

Here we have two `Drawable` objects: `CirclePointRadius` and `Point`, that have been drawn on the image with different colours.

* Addition of `RegularPolygon <: Drawable` to draw a regular polygon on an image.
```
img = zeros(RGB, 400, 400)
square = RegularPolygon(Point(200,200), 4, 50, π/4)
triangle = RegularPolygon(Point(100,50), 3, 20, 0)
octagon = RegularPolygon(Point(200,200), 8, 50, π/4)
imshow(draw(img, [square, triangle, octagon], [RGB(0,1,0), RGB(0,0,1), RGB(1,0,0)]))
```
img2

* `Line` derived types have infinite length in contrast to `LineSegment` which are finite.
```
img = zeros(Gray, 400, 400)
l = LineTwoPoints(100, 100, 300, 300)
ls = LineSegment(300, 100, 100, 300)
imshow(draw(img, [l,ls]))
```
img3

* Some more `Drawable` types have been added like `CircleThreePoints`, `Path` and `Polygon`:
```
p1 = Point(200, 100); p2 = Point(100, 200); p3 = Point(300,200)
circle = CircleThreePoints(p1, p2, p3)
img = zeros(RGB, 400 ,400)
polygon = Polygon([p1,p2,p3])
imshow(draw(img, [circle, polygon, p1, p2, p3], [RGB(1,0,0), RGB(0,0,1)]))
```
img4

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.5
 ColorTypes 0.3.2
 ImageCore 0.1.1
-Distances
+Distances 0.1.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.5
 ColorTypes 0.3.2
 ImageCore 0.1.1
 Distances 0.1.0
+Compat 0.17.0
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.5
 ColorTypes 0.3.2
 ImageCore 0.1.1
+Distances
--- a/src/ImageDraw.jl
+++ b/src/ImageDraw.jl
@@ -1,7 +1,7 @@
 module ImageDraw
 
 # package code goes here
-using ImageCore, ColorTypes, Distances
+using ImageCore, ColorTypes, Distances, Compat
 
 include("core.jl")
 include("line2d.jl")
--- a/src/ImageDraw.jl
+++ b/src/ImageDraw.jl
@@ -1,34 +1,48 @@
 module ImageDraw
 
 # package code goes here
-using ImageCore, ColorTypes
+using ImageCore, ColorTypes, Distances
 
+include("core.jl")
 include("line2d.jl")
 include("ellipse2d.jl")
 include("circle2d.jl")
 include("paths.jl")
 
-#export Types
-export LineTwoPoints, LineNormal
-
-#export Methods
+#export methods
 export
-	#Lines
-	line!,
-	line,
+	draw,
+	draw!,
 	bresenham,
-	xiaolin_wu,
+	xiaolin_wu
+
+#export types
+export
+	#Drawable
+	Drawable,
+
+	#Point
+	Point,
+
+	#Lines
+	LineNormal,
+	LineTwoPoints,
+
+	#LineSegment
+	LineSegment,
 
 	#Ellipse
-	ellipse!,
-	ellipse,
+	Ellipse,
 
-	#Circle
-	circle!,
-	circle,
+	#Circles
+	CirclePointRadius,
+	CircleThreePoints,
 
-	#Paths
-	path!,
-	path
+	#Path
+	Path,
+
+	#Polygons
+	Polygon,
+	RegularPolygon
 
 end # module
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -18,10 +18,10 @@ function draw!{T<:Colorant}(img::AbstractArray{T, 2}, circle::CircleThreePoints,
     x2 = circle.p2.x; y2 = circle.p2.y
     x3 = circle.p3.x; y3 = circle.p3.y
     D = det([[x1, x2, x3] [y1, y2, y3] [1, 1, 1]])
-    ! isapprox(D, 0, rtol = 0, atol = eps(Float64)) || error("Points do not lie on unique circle")
+    ! isapprox(D, 0, rtol = 0, atol = maximum(abs, (x1, x2, x3)) * maximum(abs, (y1, y2, y3)) * eps(Float64)) || error("Points do not lie on unique circle")
     R = [[y2 - y3, x3 - x2] [y2 - y1, x1 - x2]] * [(x1^2 + y1^2)-(x2^2 + y2^2), (x2^2 + y2^2)-(x3^2 + y3^2)] / (2 * D)
     ρ = euclidean([x1, y1], R)
     (first(ind[2]) <= R[1] <= last(ind[2]) && first(ind[1]) <= R[2] <= last(ind[1])) || error("Center of circle is out of the bounds of image")
-    ρ - 1 <= minabs([R[1] - first(ind[2]), R[1] - last(ind[2]), R[2] - first(ind[1]), R[2] - last(ind[1])]) || error("Circle is out of the bounds of image : Radius is too large")
+    ρ - 1 <= minimum(abs, [R[1] - first(ind[2]), R[1] - last(ind[2]), R[2] - first(ind[1]), R[2] - last(ind[1])]) || error("Circle is out of the bounds of image : Radius is too large")
     draw!(img, CirclePointRadius(Point(round(Int,R[1]), round(Int,R[2])), ρ), color)
 end
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -1,15 +1,24 @@
-"""
-```
-img_with_circle = circle(img, center, radiusy, radiusx)
-img_with_circle = circle(img, center, color, radiusy, radiusx)
-img_with_circle = circle(img, y, x, radiusy, radiusx)
-img_with_circle = circle(img, y, x, color, radiusy, radiusx)
-```
-Draws a circle on the input image given the `center` as a `CartesianIndex{2}` or
-as coordinates `(y, x)` using the specified `color`. If `color` is not specified,
-`one(eltype(img))` is used.
-"""
-circle{T<:Colorant}(img::AbstractArray{T, 2}, args...) = circle!(copy(img), args...)
-circle!{T<:Colorant}(img::AbstractArray{T, 2}, center::CartesianIndex{2}, color::T, radius::Real) = circle!(img, center[1], center[2], color, radius)
-circle!{T<:Colorant}(img::AbstractArray{T, 2}, y::Int, x::Int, radius::Real) = circle!(img, y, x, one(T), radius)
-circle!{T<:Colorant}(img::AbstractArray{T, 2}, y::Int, x::Int, color::T, radius::Real) = ellipse!(img, y, x, color, radius, radius)
+#CirclePointRadius methods
+
+CirclePointRadius(x::Int, y::Int, ρ::Real) = CirclePointRadius(Point(x,y), ρ)
+CirclePointRadius(p::CartesianIndex{2}, ρ::Real) = CirclePointRadius(Point(p), ρ)
+
+draw!{T<:Colorant}(img::AbstractArray{T, 2}, circle::CirclePointRadius, color::T) = draw!(img, Ellipse(circle), color)
+
+#CircleThreePoints methods
+
+CircleThreePoints(x1::Int, y1::Int, x2::Int, y2::Int, x3::Int, y3::Int) =
+    CircleThreePoints(Point(x1,y1), Point(x2,y2), Point(x3,y3))
+CircleThreePoints(p1::CartesianIndex{2}, p2::CartesianIndex{2}, p3::CartesianIndex{2}) =
+    CircleThreePoints(Point(p1), Point(p2), Point(p3))
+
+function draw!{T<:Colorant}(img::AbstractArray{T, 2}, circle::CircleThreePoints, color::T)
+    x1 = circle.p1.x; y1 = circle.p1.y
+    x2 = circle.p2.x; y2 = circle.p2.y
+    x3 = circle.p3.x; y3 = circle.p3.y
+    D = det([[x1, x2, x3] [y1, y2, y3] [1, 1, 1]])
+    D != 0 || error("Points do not lie on unique circle")
+    R = [[y2 - y3, x3 - x2] [y2 - y1, x1 - x2]] * [(x1^2 + y1^2)-(x2^2 + y2^2), (x2^2 + y2^2)-(x3^2 + y3^2)] / (2 * D)
+    ρ = euclidean([x1, y1], R)
+    draw!(img, CirclePointRadius(Point(R[1], R[2]), ρ), color)
+end
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -13,12 +13,15 @@ CircleThreePoints(p1::CartesianIndex{2}, p2::CartesianIndex{2}, p3::CartesianInd
     CircleThreePoints(Point(p1), Point(p2), Point(p3))
 
 function draw!{T<:Colorant}(img::AbstractArray{T, 2}, circle::CircleThreePoints, color::T)
+    ind = indices(img)
     x1 = circle.p1.x; y1 = circle.p1.y
     x2 = circle.p2.x; y2 = circle.p2.y
     x3 = circle.p3.x; y3 = circle.p3.y
     D = det([[x1, x2, x3] [y1, y2, y3] [1, 1, 1]])
-    D != 0 || error("Points do not lie on unique circle")
+    ! isapprox(D, 0, rtol = 0, atol = eps(Float64)) || error("Points do not lie on unique circle")
     R = [[y2 - y3, x3 - x2] [y2 - y1, x1 - x2]] * [(x1^2 + y1^2)-(x2^2 + y2^2), (x2^2 + y2^2)-(x3^2 + y3^2)] / (2 * D)
     ρ = euclidean([x1, y1], R)
+    (first(ind[2]) <= R[1] <= last(ind[2]) && first(ind[1]) <= R[2] <= last(ind[1])) || error("Center of circle is out of the bounds of image")
+    ρ - 1 <= minabs([R[1] - first(ind[2]), R[1] - last(ind[2]), R[2] - first(ind[1]), R[2] - last(ind[1])]) || error("Circle is out of the bounds of image : Radius is too large")
     draw!(img, CirclePointRadius(Point(round(Int,R[1]), round(Int,R[2])), ρ), color)
 end
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -20,5 +20,5 @@ function draw!{T<:Colorant}(img::AbstractArray{T, 2}, circle::CircleThreePoints,
     D != 0 || error("Points do not lie on unique circle")
     R = [[y2 - y3, x3 - x2] [y2 - y1, x1 - x2]] * [(x1^2 + y1^2)-(x2^2 + y2^2), (x2^2 + y2^2)-(x3^2 + y3^2)] / (2 * D)
     ρ = euclidean([x1, y1], R)
-    draw!(img, CirclePointRadius(Point(R[1], R[2]), ρ), color)
+    draw!(img, CirclePointRadius(Point(round(Int,R[1]), round(Int,R[2])), ρ), color)
 end
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -1,7 +1,7 @@
 #CirclePointRadius methods
 
-CirclePointRadius(x::Int, y::Int, ρ::Real) = CirclePointRadius(Point(x,y), ρ)
-CirclePointRadius(p::CartesianIndex{2}, ρ::Real) = CirclePointRadius(Point(p), ρ)
+CirclePointRadius{T<:Real}(x::Int, y::Int, ρ::T) = CirclePointRadius(Point(x,y), ρ)
+CirclePointRadius{T<:Real}(p::CartesianIndex{2}, ρ::T) = CirclePointRadius(Point(p), ρ)
 
 draw!{T<:Colorant}(img::AbstractArray{T, 2}, circle::CirclePointRadius, color::T) = draw!(img, Ellipse(circle), color)
 
--- a/src/core.jl
+++ b/src/core.jl
@@ -1,0 +1,170 @@
+"""
+Type representing any object drawable on image
+"""
+abstract Drawable
+
+"""
+    p = Point(x,y)
+    p = Point(c)
+
+A `Drawable` point on the image
+"""
+immutable Point <: Drawable
+    x::Int
+    y::Int
+end
+
+abstract Line <: Drawable
+abstract Circle <: Drawable
+
+
+"""
+    line = LineTwoPoints(p1, p2)
+
+A `Drawable` infinite length line passing through the two points
+`p1` and `p2`.
+"""
+immutable LineTwoPoints <: Line
+    p1::Point
+    p2::Point
+end
+
+"""
+    line = LineNormal(ρ, θ)
+
+A `Drawable` infinte length line having perpendicular length `ρ` from
+origin and angle `θ` between the perpendicular and x-axis
+"""
+immutable LineNormal <: Line
+    ρ::Real
+    θ::Real
+end
+
+"""
+    circle = CircleThreePoints(p1, p2, p3)
+
+A `Drawable` circle passing through points `p1`, `p2` and `p3`
+"""
+immutable CircleThreePoints <: Circle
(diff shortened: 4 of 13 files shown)


src/core.jl
```diff
@@ -0,0 +1,170 @@
...
+
+"""
+    line = LineNormal(ρ, θ)
+
+A `Drawable` infinte length line having perpendicular length `ρ` from
+origin and angle `θ` between the perpendicular and x-axis
+"""
+immutable LineNormal <: Line
+    ρ::Real
+    θ::Real
```
Evizero: should we make those `Float64` or parametric?

src/core.jl
```diff
@@ -0,0 +1,170 @@
...
+
+Draws all objects in `[drawable]` in the given order on `img` using
+corresponding colors from `[color]` which defaults to `one(eltype(img))`
+"""
+function draw!{T<:Colorant, U<:Drawable, V<:Colorant}(img::AbstractArray{T,2}, objects::Vector{U}, colors::Vector{V})
+    colors = copy(colors)
+    while length(colors) < length(objects)
+        push!(colors, one(T))
+    end
+    map(objects, colors) do object, color
```
Evizero: This seems like a good example to use `foreach` instead of `map`

src/paths.jl
```diff
@@ -23,14 +15,32 @@ function path!{T<:Colorant}(img::AbstractArray{T, 2}, vertices::Array{CartesianI
...
 
-    if closed==true
-        line!(img, vertices[1], vertices[end], color)
-    end
-end
\ No newline at end of file
+#Polygon methods
+
+Polygon(v::Vector{Tuple{Int, Int}}) = Polygon([Point(p...) for p in v])
+Polygon(v::Vector{CartesianIndex{2}}) = Polygon([Point(p) for p in v])
```
Evizero: Two orthogonal comments:

- Do these result on the same as `Polygon.(v)` ? Should we provide explicit vectorized functions if broadcast would do the trick?

- If we do provide these, dispatching on `AbstractVector` would be a little more flexible

src/ImageDraw.jl
```diff
@@ -1,34 +1,48 @@
 module ImageDraw
 
 # package code goes here
-using ImageCore, ColorTypes
+using ImageCore, ColorTypes, Distances
```
Evizero: Needs an entry in `REQUIRE`

src/core.jl
annimesh2809: `Float64` might be too strict for 32-bit machines, parametric could be done.
Are we doing this because `Real` is an abstract type and causes performance issues?

Evizero: Performance, yes. Consider this toy example

```julia
julia> immutable Point
           x::Real
       end

julia> f(p, i) = p.x * i

julia> p = Point(0.7)
Point(0.7)

julia> @code_llvm f(p, 0.7)

define %jl_value_t* @julia_f_72039(%jl_value_t*, double) #0 {
top:
  %ptls_i8 = call i8* asm "movq %fs:0, $0;\0Aaddq $$-2672, $0", "=r,~{dirflag},~{fpsr},~{flags}"() #2
  %ptls = bitcast i8* %ptls_i8 to %jl_value_t***
  %2 = alloca [5 x %jl_value_t*], align 8
  %.sub = getelementptr inbounds [5 x %jl_value_t*], [5 x %jl_value_t*]* %2, i64 0, i64 0
  %3 = getelementptr [5 x %jl_value_t*], [5 x %jl_value_t*]* %2, i64 0, i64 2
  %4 = bitcast %jl_value_t** %3 to i8*
  call void @llvm.memset.p0i8.i32(i8* %4, i8 0, i32 24, i32 8, i1 false)
  %5 = bitcast [5 x %jl_value_t*]* %2 to i64*
  store i64 6, i64* %5, align 8
  %6 = getelementptr [5 x %jl_value_t*], [5 x %jl_value_t*]* %2, i64 0, i64 1
  %7 = bitcast i8* %ptls_i8 to i64*
  %8 = load i64, i64* %7, align 8
  %9 = bitcast %jl_value_t** %6 to i64*
  store i64 %8, i64* %9, align 8
  store %jl_value_t** %.sub, %jl_value_t*** %ptls, align 8
  %10 = getelementptr [5 x %jl_value_t*], [5 x %jl_value_t*]* %2, i64 0, i64 4
  %11 = getelementptr [5 x %jl_value_t*], [5 x %jl_value_t*]* %2, i64 0, i64 3
  %12 = bitcast %jl_value_t* %0 to i64*
  %13 = load i64, i64* %12, align 8
  store %jl_value_t* inttoptr (i64 140216341607648 to %jl_value_t*), %jl_value_t** %3, align 8
  %14 = bitcast %jl_value_t** %11 to i64*
  store i64 %13, i64* %14, align 8
  %15 = call %jl_value_t* @jl_gc_pool_alloc(i8* %ptls_i8, i32 1432, i32 16)
  %16 = getelementptr inbounds %jl_value_t, %jl_value_t* %15, i64 -1, i32 0
  store %jl_value_t* inttoptr (i64 140216337464912 to %jl_value_t*), %jl_value_t** %16, align 8
  %17 = bitcast %jl_value_t* %15 to double*
  store double %1, double* %17, align 8
  store %jl_value_t* %15, %jl_value_t** %10, align 8
  %18 = call %jl_value_t* @jl_apply_generic(%jl_value_t** %3, i32 3)
  %19 = load i64, i64* %9, align 8
  store i64 %19, i64* %7, align 8
  ret %jl_value_t* %18
}
```

You will see the output of the LLVM code involved the garbage collector

```
  %15 = call %jl_value_t* @jl_gc_pool_alloc(i8* %ptls_i8, i32 1432, i32 16)
```

and use dynamic dispatch

```
  %18 = call %jl_value_t* @jl_apply_generic(%jl_value_t** %3, i32 3)
```

While the following is a lot simpler

```julia
julia> immutable Pointf
           x::Float64
       end

julia> p = Pointf(0.7)
Pointf(0.7)

julia> @code_llvm f(p, 0.7)

define double @julia_f_72066(%Pointf*, double) #0 {
top:
  %2 = getelementptr inbounds %Pointf, %Pointf* %0, i64 0, i32 0
  %3 = load double, double* %2, align 8
  %4 = fmul double %3, %1
  ret double %4
}
```

src/paths.jl
annimesh2809: > Do these result in the same as `Polygon.(v)` ? Should we provide explicit vectorized methods if broadcast would do the trick?

I think you misinterpreted these methods. `Polygon` requires a list of `Point` as a parameter. That list can be passed as either [`Vector{Point}`](https://github.com/JuliaImages/ImageDraw.jl/pull/13/files/94c38d84e47b81d61046c1c87bc3f1072d6029d3#diff-ac236c9f23e3c15683602f303d89d8cdR100) , [`Vector{Tuple{Int, Int}}`](https://github.com/JuliaImages/ImageDraw.jl/pull/13/files/94c38d84e47b81d61046c1c87bc3f1072d6029d3#diff-7473990870b1a3ffad8f07ae087f997dR28) or [`Vector{CartesianIndex{2}}`](https://github.com/JuliaImages/ImageDraw.jl/pull/13/files/94c38d84e47b81d610946c1c87bc3f1072d6029d3#diff-7473990870b1a3ffad8f07ae087f97dR29). A `Polygon` cannot be represented using just a single point so `Polygon.(v)` won't work.

> If we do provide these, dispatching on `AbstractVector` would be a little more flexible

Yes, that would be better :smile:

Evizero: > I think you misinterpreted these methods. 

yes, you are absolutely right.

src/core.jl
annimesh2809: Yes, indeed it gave about 10 times performance boost (mean time). I will do the same for `Ellipse`, `CirclePointRadius` etc.
```julia
julia> immutable LineNormal
           ρ::Real
           θ::Real
       end

julia> immutable LineNormalParam{T<:Real, U<:Real}
           ρ::T
           θ::U
       end

julia> @benchmark LineNormal(1.0,1.0)
BenchmarkTools.Trial: 
  memory estimate:  32 bytes
  allocs estimate:  1
  --------------
  minimum time:     8.813 ns (0.00% GC)
  median time:      11.701 ns (0.00% GC)
  mean time:        13.548 ns (14.39% GC)
  maximum time:     1.697 μs (98.15% GC)
  --------------
  samples:          10000
  evals/sample:     999
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark LineNormalParam(1.0,1.0)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.888 ns (0.00% GC)
  median time:      1.892 ns (0.00% GC)
  mean time:        1.984 ns (0.00% GC)
  maximum time:     539.526 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
  time tolerance:   5.00%
  memory tolerance: 1.00%
```
But why doesn't `@code_warntype` give any information about it?

```julia
julia> @code_warntype LineNormal(1.0,1.0)
Variables:
  #self#::Type{LineNormal}
  ρ::Float64
  θ::Float64

Body:
  begin 
      return $(Expr(:new, :(Main.LineNormal), :(ρ), :(θ)))
  end::LineNormal
```

Evizero: Well, the thing you are testing (i.e. the constructor) isn't really the bottleneck I was thinking off. I am more worried about the algorithm that use those variables in a loop multiple times. I'll point one such position out to you with a separate comment.

src/ellipse2d.jl
```diff
@@ -1,30 +1,15 @@
...
-function ellipse!{T<:Colorant}(img::AbstractArray{T, 2}, y::Int, x::Int, color::T, radiusy::Real, radiusx::Real)
+function draw!{T<:Colorant}(img::AbstractArray{T, 2}, ellipse::Ellipse, color::T)
 	ys = Int[]
 	xs = Int[]
-	for i in y : y + radiusy
-		for j in x : x + radiusx
-			val = ((i - y) / radiusy) ^ 2 + ((j - x) / radiusx) ^ 2
+	for i in ellipse.center.y : ellipse.center.y + ellipse.ρy
+		for j in ellipse.center.x : ellipse.center.x + ellipse.ρx
+			val = ((i - ellipse.center.y) / ellipse.ρy) ^ 2 + ((j - ellipse.center.x) / ellipse.ρx) ^ 2
```
Evizero: Concerning performance bottleneck and parameteric types (please @timholy correct me here if I am mistaken):

Consider this nested loop. If the member variables of the variable `ellipse` are of abstract types, then two things are true. 

- The member variables of `ellipse` are located on the heap, not the stack, regardless if the type `Ellipse` itself is `immutable`. 

- The loop variables `i` and `j` are not inferrable, which means that every function in the inner loop that uses those is invoked using dynamic dispatch (runtime method lookup). Dynamic dispatch itself isn't a problem per say (it is a useful option to have), but it is undesirable to have dynamic dispatch happen inside a (nested) loop, because that code is potentially executed many times.

src/core.jl
Evizero: The performance difference you are observing is probably mostly due to the fact that the member variables of `LineNormal` need to be allocated on the heap. This is because those member variables are declared as abstract types. Consequently the compiler doesn't know beforehand how much space they need. 

Concerning your `@code_warntype` question: The parameters to the constructor are inferrable and so is the return type. That means there is no inference problem to report and `@code_warntype` is doing it's job.  However, once you have a function that uses the member variables, which are not inferable, then you'll see the red `Any` pop up as a consequence:

```julia
julia> immutable Point
           x::Real
       end

julia> f(p, i) = p.x * i
f (generic function with 1 method)

julia> p = Point(0.7)
Point(0.7)

julia> @code_warntype f(p, 0.7)
Variables:
  #self#::#f
  p::Point
  i::Float64

Body:
  begin 
      return ((Core.getfield)(p::Point,:x)::Real * i::Float64)::Any
  end::Any
```

src/core.jl
```diff
@@ -0,0 +1,170 @@
...
+Draws the `drawable` object on a copy of image `img` using color
+`color`. Can also draw multiple `Drawable` objects when passed
+as a `Vector{Drawable}` with corresponding colors in `[color]` 
+"""
+draw{T<:Colorant}(img::AbstractArray{T,2}, args...) = draw!(copy(img), args...)
+
+Point(p::CartesianIndex) = Point(p[2], p[1])
+
+function draw!{T<:Colorant}(img::AbstractArray{T,2}, point::Point, color::T)
+    if CartesianIndex(point.y, point.x) ∈ CartesianRange(size(img))
```
timholy: `if checkbounds(Bool, img, point.y, point.x)` is an alternative. If you keep this version, please change `size` to `indices`.

src/core.jl
```diff
@@ -0,0 +1,170 @@
...
+A `Drawable` circle passing through points `p1`, `p2` and `p3`
+"""
+immutable CircleThreePoints <: Circle
+    p1::Point
+    p2::Point
+    p3::Point
+end
+
+"""
+    circle = CirclePointRadius(p, ρ)
```
timholy: Just for readability I might change this to `CirclePointRadius(c, ρ)` or `CirclePointRadius(center, ρ)`. The difference between `p` and `ρ` might be subtle.

src/core.jl
```diff
@@ -0,0 +1,170 @@
...
+immutable Ellipse <: Drawable
+    center::Point
+    ρx::Real
+    ρy::Real
+end
+
+"""
+    polygon = Polygon([vertex])
+
+A `Drawable` polygon represented by a `Vector` of vertices
```
timholy: What's the difference between a Path and a Polygon? Assuming it's closure, would be good to mention.

src/line2d.jl
```diff
@@ -1,84 +1,73 @@
...
-    x1::Number
-end
-
-LineTwoPoints(p1::CartesianIndex{2}, p2::CartesianIndex{2}) = LineTwoPoints(p1[1],p1[2],p2[1],p2[2])
+#Function to return valid intersections of lines with image boundary
 
-immutable LineNormal <: Line
-    ρ::Number
-    θ::Number
+function get_valid_intersections(intersections::Vector{Tuple{Real, Real}}, indsx::Base.OneTo{Int}, indsy::Base.OneTo{Int})
```
timholy: signature on `indsx` and `indsy` is too restrictive, just make them `AbstractUnitRange` or something. Note that `first(::OneTo) == 1`, so there was a mismatch of generality here.

src/line2d.jl
```diff
@@ -1,84 +1,73 @@
...
-    line!(img, LineTwoPoints(p1, p2), one(T), method)
-
-line!{T<:Colorant}(img::AbstractArray{T, 2}, p1::CartesianIndex{2}, p2::CartesianIndex{2}, color::T, method::Function = bresenham) =
-    line!(img, LineTwoPoints(p1, p2), color, method)
+function draw!{T<:Colorant}(img::AbstractArray{T,2}, line::LineTwoPoints, color::T, method::Function = bresenham)
+    indsy, indsx = indices(img)
+    x1 = line.p1.x; y1 = line.p1.y
+    x2 = line.p2.x; y2 = line.p2.y
+    m = (y2 - y1) / (x2 - x1)
+    intersections_x = [(x, y1 + m*(x-x1)) for x in (first(indsx), last(indsx))]
```
timholy: What will happen for a vertical line?

src/core.jl
annimesh2809: Thanks for the clear explanation. :smile:

src/line2d.jl
annimesh2809: In case of vertical line `m = Inf`, so `intersections_x = [(first(indsx),Inf), (last(indsx),Inf)]` and `intersections_y = [(x1, first(indsy)),(x1, last(indsy)]`. `valid_intersections = [(x1, first(indsy)),(x1, last(indsy)]` and so the line would be drawn out correctly.
```julia
julia> imshow(draw(zeros(Gray,400,400), LineTwoPoints(200,10,200,20)))
```
img5

annimesh2809:
I wanted to make `LineNormal` parametric but it would require two parameters:
```julia
immutable LineNormal{T<:Real,U<:Real} <: Line
    ρ::T
    θ::U
end
```
So I thought it would be better to make them both Float. But unlike `Int` which is a `typealias` for system's native integer type, I could not find any native `typealias` for float. I can write an alias like:
```julia
if Sys.WORD_SIZE == 64
    typealias Float Float64
else
    typealias Float Float32
end
```
Should I do it this way? Or make it parametric? Or is there some other nice way to do this?

Evizero:
http://stackoverflow.com/a/21746036

timholy:
Presumably it would be fine to insist that they both have the same concrete type?

Evizero:
> Presumably it would be fine to insist that they both have the same concrete type? 

I also think that is a good idea. I'd be reluctant to hard code `Float64` or `Float32` because a lot of images work with different numbers internally. I could imagine other code that uses this library wanting to create `LineNormal` etc with types based on the concrete image `eltype` one is working with. Having an option is good.

tldr, something like 

```julia
mmutable LineNormal{T<:Number} <: Line
    ρ::T
    θ::T
end
```

timholy:
From a quick scan, this seems ready to merge. Let me know when you want me to pull the trigger. (Of course, anyone who spots issues should not hesitate to mention them.)

annimesh2809:
Let's first add some more tests to cover up `src/ellipse2d.jl` and `src/circle2d.jl`. I tried to cover up `src/paths.jl` completely but the lines [20, 21](https://github.com/JuliaImages/ImageDraw.jl/blob/master/src/paths.jl#L21), [28 and 29](https://github.com/JuliaImages/ImageDraw.jl/blob/master/src/paths.jl#L29) could not be covered due to `error` function. Should we leave those lines uncovered?

src/core.jl
```diff
@@ -0,0 +1,180 @@
...
+    ρx::T
+    ρy::U
+end
+
+"""
+    polygon = Polygon([vertex])
+
+A `Drawable` polygon i.e. a closed path created by joining the
+consecutive points in `[vertex]` along with the first and last point.
+!!! note
```
Evizero: Does this doc-string render properly in the REPL?  I think there needs to be a newline above this line

test/core.jl
```diff
@@ -0,0 +1,38 @@
+@testset "Core" begin
+
+    points = [Point(1,1), Point(2,2), Point(3,3)]
+    img = zeros(Gray{N0f8}, 3, 3)
+    draw!(img, points)
```
Evizero: we may want to change these tests to 

```julia
@inferred draw!(img, points)
```

in the future to spot obvious inference issues. I don't think this needs to be done right now though

test/line2d.jl
```diff
@@ -1,32 +1,32 @@
 @testset "Lines2D" begin
 
-	@testset "Lines API" begin
+	@testset "LineSegment" begin
 		img = zeros(Gray{N0f8}, 10, 10)
 		expected = copy(img)
 		expected[diagind(expected)] = one(Gray{N0f8})
-		res = line(img, 1, 1, 10, 10)
+		res = draw(img, LineSegment(1,1,10,10))
```
Evizero: same here with

```julia
res = @inferred draw(img, LineSegment(1,1,10,10))
```

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.5
 ColorTypes 0.3.2
 ImageCore 0.1.1
+Distances
```
Evizero: Tony will want us to be more specific with the minimum version here

Evizero:
> but the lines 20, 21, 28 and 29 could not be covered due to error function.

Unsure what you are saying, but in case you did not know, you can test if a function throws an error

```julia
julia> f(x) = error()
f (generic function with 1 method)

julia> @test_throws ErrorException f(2)
Test Passed
  Expression: f(2)
      Thrown: ErrorException
```

annimesh2809:
Wow, I didn't know we can even test for function throwing an error. Thanks @Evizero :)

src/core.jl
annimesh2809: It seems to be rendering same with or without the newline.

timholy:
To figure out the 32-bit error, you might need to push a "test" commit that does something like
```julia
D = det([1 2 3; 1 2 3; 1 1 1])
@show D
```
and then (if it makes it that far) calculates the two components of `R` and shows them. The `InexactError` might be coming from the `round(Int,R[i])` statements if `D` turns out not to be exactly 0 but the value of `abs(R[i])` is bigger than `typemax(Int32)` (or is NaN).

annimesh2809:
> The InexactError might be coming from the round(Int,R[i]) statements if D turns out not to be exactly 0 but the value of abs(R[i]) is bigger than typemax(Int32) (or is NaN).

You are absolutely correct. I tested it out on another branch and here is the result:
imgt

Seems like `D` is not exactly 0. We can fix that by changing the [test condition](https://github.com/JuliaImages/ImageDraw.jl/pull/13/files#diff-b1fdba638d243809d21e263ece7140f8R20) to 
`! isapprox(D, 0, rtol = 0, atol = eps(Float64)) || error("Points do not lie on unique circle")`.  
However, the fundamental problem here is that `round` cannot fit data larger than `typemax(Int32)` into a block of size `Int32`. For normal usage, the radius and center of the circle will be way smaller than these bounds. To cover the unsound cases, we can put a limit on the allowed radius and center of the circle to be drawn else throw an error:
```julia
ind = indices(img)
(first(ind[2]) <= R[1] <= last(ind[2]) && first(ind[1]) <= R[2] <= last(ind[1])) || error("center not in image")
ρ <= minabs([R[1] - first(ind[2]), R[1] - last(ind[2]), R[2] - first(ind[1]), R[2] - last(ind[1])])
|| error("radius is too large")
```
This way, before reaching the `round` part, the condition would fail and throw an error.

src/circle2d.jl
```diff
@@ -13,12 +13,15 @@ CircleThreePoints(p1::CartesianIndex{2}, p2::CartesianIndex{2}, p3::CartesianInd
     CircleThreePoints(Point(p1), Point(p2), Point(p3))
 
 function draw!{T<:Colorant}(img::AbstractArray{T, 2}, circle::CircleThreePoints, color::T)
+    ind = indices(img)
     x1 = circle.p1.x; y1 = circle.p1.y
     x2 = circle.p2.x; y2 = circle.p2.y
     x3 = circle.p3.x; y3 = circle.p3.y
     D = det([[x1, x2, x3] [y1, y2, y3] [1, 1, 1]])
-    D != 0 || error("Points do not lie on unique circle")
+    ! isapprox(D, 0, rtol = 0, atol = eps(Float64)) || error("Points do not lie on unique circle")
```
timholy: Threshold comparison can be surprisingly tricky. Here one issue is that this criterion is not "scale invariant," meaning that I can easily fool the criterion you specified just by multiplying `x1,  x2, x3` by 10. How about making the threshold `maximum(abs, (x1, x2, x3)) * maximum(abs, (y1, y2, y3)) * eps(Float64)`?

annimesh2809:
I think this PR is now ready. If someone finds any issue, do comment so.

An unrelated question regarding `REQUIRE` file:
As suggested in this [link](https://docs.julialang.org/en/release-0.5/manual/packages/#man-package-requirements), we should specify the names of the packages along with the range of versions that supports features required by our package. So how do we go about looking for the version range for any package that supports the required feature? Currently, I had to do a binary search on the tags to find the minimum version ([v0.17.0](https://github.com/JuliaLang/Compat.jl/tree/v0.17.0)) of `Compat.jl` that supported `@compat abstract type T end`. Is there a better way like a features log that logs the version number along with features added in that version?

timholy:
Good job finding the proper Compat version. I'm not certain we need version bounds on all packages, but Compat is especially important.

timholy:
Thank you, @annimesh2809! This really matures this package considerably.

Evizero:
🎉

annimesh2809:
Thanks to both of you for all your help and suggestions! :smile:

PR #21, opened 2018-09-17
Fix for v0.7/1.0

Dsuke-K:
indices -> axes


timholy:
For the appveyor failures, you can likely just replace the current file with [the one in the README](https://github.com/juliaci/appveyor.jl).

PR #23, opened 2019-04-30
Switch to new registrator and reduce package dependency

johnnychen94:
* Add Project.toml and delete REQUIRE (https://github.com/JuliaImages/Images.jl/issues/798)
* remove ColorTypes and Compat dependency (https://github.com/JuliaImages/Images.jl/issues/802)
* upgrade travis.yml and drop julia v0.7 support

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,7 +9,7 @@ ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
 [compat]
 Distances = ">= 0.1.0"
 ImageCore = ">= 0.8.1"
-julia = ">= 1.0"
+julia = "1"
 
 [extras]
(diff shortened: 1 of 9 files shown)


Project.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.2.0"
+
+[deps]
+Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+ImageCore = "a09fc81d-aa75-5fe9-8630-4744c3626534"
+
+[compat]
+Distances = ">= 0.1.0"
+ImageCore = ">= 0.8.1"
+julia = ">= 1.0"
```
timholy: As I've learned myself recently, it may be better to use `julia = "1"`. The best guess is that Julia 2.0.0 will break this package in some way until it gets modified for whatever language changed that introduces, so better to declare incompatibility right away.

PR #25, opened 2019-05-17
Create cross.jl

### LINKED ISSUE #24: Add a cross
I often need to mark something in the image without obscuring it so much. A cross does exactly that. 

I'll therefor try to PR something like this:
```julia
struct Cross <: Drawable
    c::Point
    arm::Int
end

function draw!(img::AbstractArray{T, 2}, cross::Cross, color::T) where T<:Colorant
    for x in -cross.arm:cross.arm
        img[cross.c.y, cross.c.x + x] = color
    end
    for y in -cross.arm:cross.arm
        img[cross.c.y + y, cross.c.x] = color
    end
    img
end
```

yakir12:
Fixes #24

### DIFF
--- a/src/ImageDraw.jl
+++ b/src/ImageDraw.jl
@@ -8,6 +8,7 @@ include("line2d.jl")
 include("ellipse2d.jl")
 include("circle2d.jl")
 include("paths.jl")
+include("cross.jl")
 
 #export methods
 export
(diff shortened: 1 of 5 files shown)


timholy:
This seems fine to me. Perhaps the only question I have is whether this should be a subtype of `Marker` or `LineCollection` (neither of which exists yet) or something like that. But perhaps we could wait for someone to want that generalization.

I'll merge now, given the absence of other comments here.

PR #26, opened 2019-06-03
Added checkbounds() 

Codyk12:
Added checkbounds() testing to all draw!() functions as to not throw errors when shapes which are partially in the images are attempted to be drawn.

### DIFF
--- a/src/paths.jl
+++ b/src/paths.jl
@@ -4,33 +4,36 @@ Path(v::AbstractVector{Tuple{Int, Int}}) = Path([Point(p...) for p in v])
 Path(v::AbstractVector{CartesianIndex{2}}) = Path([Point(p) for p in v])
 
 function draw!(img::AbstractArray{T, 2}, path::Path, color::T) where T<:Colorant
-    vertices = [CartesianIndex(p.y, p.x) for p in path.vertices]
+	    vertices = [CartesianIndex(p.y, p.x) for p in path.vertices]
+	    f = CartesianIndex(map(r->first(r)-1, axes(img)))
+	    l = CartesianIndex(map(r->last(r), axes(img)))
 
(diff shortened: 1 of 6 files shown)


Codyk12:
I believe this functionality could be argued either way. Do we want the path functionality to throw an error or draw what it can on the image? I believe drawing what it could on the image would be beneficial to being able to draw polygons partially on the image.

src/paths.jl
```diff
@@ -4,24 +4,36 @@ Path(v::AbstractVector{Tuple{Int, Int}}) = Path([Point(p...) for p in v])
...
-    for i in 1:length(vertices)-1
-        if min(f,vertices[i+1])==f && max(l,vertices[i+1])==l
-            draw!(img, LineSegment(vertices[i], vertices[i+1]), color)
-        else
-            println(vertices[i+1])
-            error("Point coordinates out of range.")
-        end
-    end
-end
+	    vertices = [CartesianIndex(p.y, p.x) for p in path.vertices]
```
timholy: Indentation here is weird

src/paths.jl
```diff
@@ -4,24 +4,36 @@ Path(v::AbstractVector{Tuple{Int, Int}}) = Path([Point(p...) for p in v])
...
+	    f = CartesianIndex(map(r->first(r)-1, axes(img)))
+	    l = CartesianIndex(map(r->last(r), axes(img)))
+
+	    for i in 1:length(vertices)-1
+			draw!(img, LineSegment(vertices[i], vertices[i+1]), color)
+	    end
+
+		img
+	    # No need for corrdinate checking at this level as draw!(::LineSegment, ...)
```
timholy: Yep. Seems fine to just delete all this.

PR #27, opened 2019-06-04
added LinearAlgebra dependency

yakir12:
because of [this line](https://github.com/JuliaImages/ImageDraw.jl/blob/022f1aaf305fc088c48a9b01aca859002b450d5b/src/circle2d.jl#L1).


timholy:
I will tag this when #26 merges. If that doesn't happen soon feel free to ping this.

PR #29, opened 2019-06-19
Shape hierarchy

Codyk12:
Expanded out the hierarchy for the drawable objects.

PR #31, opened 2019-06-25
Thickness

Codyk12:
Added the ability to draw lines with a thickness


src/circle2d.jl
```diff
@@ -14,7 +15,7 @@ CircleThreePoints(x1::Int, y1::Int, x2::Int, y2::Int, x3::Int, y3::Int) =
 CircleThreePoints(p1::CartesianIndex{2}, p2::CartesianIndex{2}, p3::CartesianIndex{2}) =
     CircleThreePoints(Point(p1), Point(p2), Point(p3))
 
-function draw!(img::AbstractArray{T, 2}, circle::CircleThreePoints, color::T) where T<:Colorant
+function draw!(img::AbstractArray{T, 2}, circle::CircleThreePoints, color::T; in_bounds::Bool=false, thickness::Integer=-1) where T<:Colorant
```
johnnychen94: I think we can use
```
thickness::Union{Integer, Nothing}=nothing
```

This representation has more explicit meaning compared to `-1` :D

src/core.jl
```diff
@@ -156,12 +202,13 @@ corresponding colors from `[color]` which defaults to `oneunit(eltype(img))`
 If only a single color `color` is specified then all objects will be
 colored with that color.
```
johnnychen94: Perhaps you need to update the docstring for `in_bounds` and `thickness`.

src/core.jl
```diff
@@ -156,12 +202,13 @@ corresponding colors from `[color]` which defaults to `oneunit(eltype(img))`
...
 """
-function draw!(img::AbstractArray{T,2}, objects::AbstractVector{U}, colors::AbstractVector{V}) where {T<:Colorant, U<:Drawable, V<:Colorant}
+function draw!(img::AbstractArray{T,2}, objects::AbstractVector{U}, colors::AbstractVector{V}; in_bounds::AbstractVector{Bool}=[false], thickness::AbstractVector{<:Integer}=[-1]) where {T<:Colorant, U<:Drawable, V<:Colorant}
     colors = copy(colors)
-    while length(colors) < length(objects)
-        push!(colors, oneunit(T))
-    end
-    foreach((object, color) -> draw!(img, object, color), objects, colors)
+    while length(colors) < length(objects) push!(colors, oneunit(T)) end
```
johnnychen94: > If only a single color `color` is specified then all objects will be
colored with that color.	colored with that color.

The current codes don't represent this logic well.

```julia
function draw!(img::AbstractArray{T, 2}, objects::AbstractVector{U}, color::Colorant; in_bounds::Bool=false, thickness::Integer=-1) where {T<:Colorant, U<:Drawable}
    draw!(img, objects, repeat([color], length(objects));
          in_bounds = repeat([in_bounds], length(objects)), 
          thinckness = repeat([thickness], length(objects)))
end
function draw!(img::AbstractArray{T,2}, objects::AbstractVector{U}, colors::AbstractVector{V}; in_bounds::AbstractVector{Bool}, thickness::AbstractVector{<:Integer}) where {T<:Colorant, U<:Drawable, V<:Colorant}
    length(colors) == length(objects) || throw("The number of colors and objects should be equal.")
    length(in_bounds) == length(objects) || throw("")
    length(thickness) == length(objects) || throw("")
    foreach((object, color, in_b, thick) -> draw!(img, object, color), objects, colors, in_bounds, thickness)
    img
end
```

src/cross.jl
```diff
@@ -4,12 +4,12 @@ A `Drawable` cross passing through the point `c` with arms that are `arm` pixels
 """
 Cross(c, arm::Int) = Cross(c, -arm:arm)
 
-function draw!(img::AbstractArray{T, 2}, cross::Cross, color::T) where T<:Colorant
+function draw!(img::AbstractArray{T, 2}, cross::Cross, color::T; in_bounds::Bool=false, thickness::Int=-1) where T<:Colorant
```
johnnychen94:
```diff
-draw!(img::AbstractArray{T, 2}, cross::Cross, color::T; in_bounds::Bool=false, thickness::Int=-1)
+draw!(img::AbstractArray{T, 2}, cross::Cross, color::T; in_bounds::Bool=false, thickness::Integer=-1)
```

johnnychen94:
To fix CI error on 1.0:

```julia
# src/compat.jl
if v"1.0" <= VERSION < v"1.1"
    isnothing(x) = x===nothing
end
```

src/core.jl
```diff
@@ -241,7 +242,7 @@ draw!(img::AbstractArray{T,2}, p::CartesianIndex{2}, color::T = oneunit(T); in_b
 
 function draw!(img::AbstractArray{T,2}, y::Integer, x::Integer, color::T; in_bounds::Bool=false, thickness::Union{Integer, Nothing}=nothing) where T<:Colorant
     in_bounds ? img[point.y, point.x] = color : drawifinbounds!(img, y, x, color)
-    isnothing(thickness) || drawwiththickness!(img, y, x, color, in_bounds, thickness)
+    (isnothing(thickness) && thickness != 1) || drawwiththickness!(img, y, x, color, in_bounds, thickness)
```
johnnychen94: If I understand it right, the pipeline is:

1. draw a thickness=1 pixel
2. if thickness !=1, draw another pixel with thickness parameter.

We don't really need to do step 1 if `thickness!=1`, right?

The right expression is `isnothing(thickness) || thickness ==1 || drawwiththickness!(...)`

It becomes a little tricky to understand, so I'd recommend to rewrite it:
```julia
function draw!(img::AbstractArray{T,2}, y::Integer, x::Integer, color::T; in_bounds::Bool=false, thickness::Union{Integer, Nothing}=nothing) where T<:Colorant
    if !isnothing(thickness)
        drawwiththickness!(img, y, x, color, in_bounds, thickness) # recursively call drawifinbounds!
    else
        in_bounds ? img[point.y, point.x] = color : drawifinbounds!(img, y, x, color)
    end
end

function drawwiththickness!(img::AbstractArray{T,2}, y0::Int, x0::Int, color::T, in_bounds::Bool, thickness::Int) where {T<:Colorant}
    if thickness == 1
        draw!(img, y0, x0, color, in_bounds=in_bounds, thickness=nothing)
    else
        ...
    end
end
```

src/core.jl
```diff
@@ -278,15 +279,12 @@ drawwiththickness!(img::AbstractArray{T,2}, p::Point, color::T, in_bounds::Bool,
...
-    n = Int(round(thickness / 2))
-    evn = thickness % 2 == 1 ? 0 : 1
-    pixels = [i for i = -(n-evn):n]
+    m = ceil(Int, thickness/2) - 1
+    n = thickness % 2 == 0 ? m+1 : m
+    pixels = [i for i = -m:n]
 
-    for (x,y) in Combinatorics.combinations(pixels, 2)
+    for x in pixels, y in pixels
         draw!(img, y0+y, x0+x, color, in_bounds=in_bounds)
```
johnnychen94: Will it be possible when `thickness` is large enough, `in_bounds` should be `false`? Currently, if you pass `in_bounds=true`, it would always be `true`.

test/core.jl
```diff
@@ -37,4 +37,19 @@ using LinearAlgebra
     @inferred draw!(img, point)
     expected = zeros(Gray{N0f8})
     @test all(img .== expected) == true
+
+    @testset "Thickness" begin
+        point = Point(2,2)
+        img = zeros(Gray{N0f8}, 3, 3)
+        @inferred draw!(img, point, thickness=2)
```
johnnychen94: Can you add one more simple test on `CartesianIndex{2}` here, to guarantee that `CartesianIndex` works the same as `Point`.

Thanks.

test/line2d.jl
```diff
@@ -28,6 +28,18 @@
 		expected[diagind(expected)] .= RGB{N0f8}(0.2, 0.3, 0.4)
 		res = @inferred draw(img2, LineSegment(1,1,10,10), RGB{N0f8}(0.2,0.3,0.4))
 		@test expected == res
+
+	end
+
+	@testset "Thickness" begin
+		img = zeros(Gray{N0f8}, 5, 5)
+		expected = Gray{N0f8}.([1 1 0 0 0
+								1 1 1 0 0
```
johnnychen94: A mistake in indentation?

src/core.jl
```diff
@@ -200,6 +263,32 @@ drawifinbounds!(img::AbstractArray{T,2}, p::Point, color::T = oneunit(T)) where
...
+Draws pixel with given thickness
+Color Defaults to oneunit(T)
+Thickness defaults to 1
+
+"""
+
+drawwiththickness!(img::AbstractArray{T,2}, p::Point, color::T, in_bounds::Bool, thickness::Integer) where {T<:Colorant} = drawwiththickness!(img, p.y, p.x, color, in_bounds, thickness)
+drawwiththickness!(img::AbstractArray{T,2}, p::CartesianIndex{2}, color::T, in_bounds::Bool, thickness::Integer) where {T<:Colorant} = drawifinbounds!(img, Point(p), color, in_bounds, thickness)
+
+function drawwiththickness!(img::AbstractArray{T,2}, y0::Int, x0::Int, color::T, in_bounds::Bool, thickness::Int) where {T<:Colorant}
```
johnnychen94:
```diff
- thickness::Int
+ thickness::Integer
```
:D

src/core.jl
johnnychen94: What I'm worry about is when `in_bounds = true`, it will hit `OutOfBoundary` error for large
`(x, y)` in

```julia
for x in pixels, y in pixels
    draw!(img, y0+y, x0+x, color, in_bounds=in_bounds)
```

johnnychen94: 🤔 I see, but perhaps a warning is more friendly than error? I'm not very sure of it.

This doesn't belong to this PR, we can rework `drawinbounds!` in the future if there's a need.

johnnychen94: >  let it draw objects even if it goes off the image.

We can define some boundary methods as types and pass it to `draw!`, just like what `ImageFiltering` does to `imfilter`.

I'm totally fine with the current design when ImageDraw isn't fully functional enough. But in the future, we need to bring this back to the table.

src/core.jl
```diff
@@ -131,42 +137,61 @@ end
...
 end
 
 """
-    img = draw!(img, drawable, color)
-    img = draw!(img, drawable)
+    img_map = ImageMap(img, x, y)
+
+Overlays an image on an larger image. (x, y) is the center for placing the image
+"""
+struct ImageMap <: Drawable
```
johnnychen94: What's the usage of this? I can't find anywhere this struct is used in your PR.

agarret7:
Are there any plans for merging this soon? My use case relies on thickness of lines and otherwise I'll have to fallback to using PIL with PyCall.

johnnychen94:
We need to merge #29 first and rebase this PR on the top of that

I'll see if that one is good to merge.

Codyk12:
Unfortunately, this library doesn't have the biggest community supporting it, so we decided to try to use another drawing library to use in the library we're making, hopefully, these features get put merged in soon though.

ashwani-rathee:
@Codyk12 can I give it a go?I think I can do it in a more simplified way with RectanglePoints #52

PR #32, opened 2019-07-02
Canvas creation

Codyk12:
Added a canvas creation function. Allows for creating images with a specified background. Currently Solid and a striped background


src/background.jl
```diff
@@ -0,0 +1,64 @@
+"""
+    gen_img = generatecanvas(colortype, x, y)
+    gen_img = generatecanvas(colortype, x, y, background)
+    gen_img = generatecanvas(colortype, [x,y])
+    gen_img = generatecanvas(colortype, [x,y], background)
+
+    Creates an image of with the given colortype and size.
+    Defaults to a black solid image
+"""
+generatecanvas(colortype::Type, x::Int, y::Int) = zeros(colortype, (x,y))
```
timholy: Since the work is being done by `zeros`, why do we effectively have to create an alias to it?

src/background.jl
```diff
@@ -0,0 +1,64 @@
...
+    draw!(zeros(colortype, (size...)), b)
+end
+
+
+"""
+    new_img = draw!(img, solid_background)
+
+    Paints the entire given image a solid color
+"""
+function draw!(img::AbstractArray{T,2}, b::SolidBackground) where {T<:Colorant}
```
timholy: Again seems redundant with standard names:
```julia
julia> A = fill(colorant"red", 2, 2)
2×2 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}:
 RGB{N0f8}(1.0,0.0,0.0)  RGB{N0f8}(1.0,0.0,0.0)
 RGB{N0f8}(1.0,0.0,0.0)  RGB{N0f8}(1.0,0.0,0.0)

julia> fill!(A, colorant"blue")
2×2 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}:
 RGB{N0f8}(0.0,0.0,1.0)  RGB{N0f8}(0.0,0.0,1.0)
 RGB{N0f8}(0.0,0.0,1.0)  RGB{N0f8}(0.0,0.0,1.0)
```

src/background.jl
```diff
@@ -0,0 +1,64 @@
...
+    end
+    img
+end
+
+"""
+    new_img = draw!(img, striped_background)
+
+    Layers colors onto a given image to set a "background"
+"""
+function draw!(img::AbstractArray{T,2}, b::StripedBackground) where {T<:Colorant}
```
timholy: This one is interesting and adds novel functionality. However, you seem to implicitly assume that `img` is an all-zeros array, in which case maybe this shouldn't take `img` as an argument? Should we have a function called `stripes`?

src/background.jl
```diff
@@ -0,0 +1,64 @@
...
+        draw!(img, LineNormal(d,b.θ), c)
+        for i = 1:y
+            cnt = 0
+            for j = 1:x
+                if img[i,j] == zero(typeof(c))
+                    draw!(img, Point(j,i), c)
+                    cnt += 1
+                elseif img[i,j] == c
+                    break
+                elseif img[i,j] in b.colors[1:e-1]
```
timholy: Slow. How about an auxiliary array that indicates whether a given pixel has already been handled?

src/background.jl
timholy: Also worth asking: how often do you use angles different from 0 and π/2? If your stripes are all rectilinear then there are easier implementations.

src/core.jl
```diff
@@ -14,8 +14,39 @@ struct Point <: Drawable
     y::Int
 end
 
-abstract type Line <: Drawable end
-abstract type Circle <: Drawable end
+abstract type AbstractPath <: Drawable end
```
timholy: I like the addition of `Abstract` to these.

src/core.jl
```diff
@@ -203,3 +233,32 @@ function drawifinbounds!(img::AbstractArray{T,2}, y::Int, x::Int, color::T) wher
     if checkbounds(Bool, img, y, x) img[y, x] = color end
     img
 end
+
+"""
```
timholy: What about adding a keyword argument to `draw!`?

src/cross.jl
```diff
@@ -4,12 +4,12 @@ A `Drawable` cross passing through the point `c` with arms that are `arm` pixels
 """
 Cross(c, arm::Int) = Cross(c, -arm:arm)
 
-function draw!(img::AbstractArray{T, 2}, cross::Cross, color::T) where T<:Colorant
+function draw!(img::AbstractArray{T, 2}, cross::Cross, color::T, thickness::Int=1) where T<:Colorant
```
timholy: I think it's better to make this a keyword argument, since we might eventually want other modifiable features.

Issue #20, opened 2018-08-11
Create documentation + website


mprat:
This library provides many useful functions! It needs a (generated and linked) documentation website.

timholy:
Would love help doing that. A good place to do it would be over at https://github.com/JuliaImages/juliaimages.github.io (see docs at https://juliaimages.org/latest/)

PR #37, opened 2020-07-07
Add docs

mgautam98:
Solves #20 

I have added few docstrings, usage examples. I will be adding more about the algorithms used to draw later.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,0 +1,53 @@
+# ImageDraw.jl Documentation
+
+A drawing package for JuliaImages
+
+```@contents
+Depth = 3
+```
+
+## introduction
+
+ImageDraw supports basic drawing on Images. You can draw points, lines, circles, ellipse and paths.  
+
+
+## Basic usage
+
+Let's start with a drawing a circle
+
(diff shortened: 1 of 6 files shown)


docs/src/index.md
```diff
@@ -0,0 +1,41 @@
...
+julia> using TestImages, ImageDraw
+
+julia> img = testimage("lighthouse");
+
+julia> draw(copy(img), Ellipse(CirclePointRadius(350,200,100)));
+
+```
+When displayed, these three images look like this:
+
+![Draw Circle](images/lighthouse_circle.png)
```
johnnychen94, line 28: The assets can be generated using `@example` block so that we don't need to *frequently* check-in these images, see [ImageEdgeDetections Adds documentation and examples](https://github.com/JuliaImages/ImageEdgeDetection.jl/pull/6/files#diff-493d89206c11fea99259d0771034d26aR56-R70) for an example.

mgautam98: Done

johnnychen94:
Do you need me to set up a democards page for this repo? Looks like it's more illustrative with it.

PR #41, opened 2020-07-22
Add rectangle shape

mgautam98:
Added rectangle shape with test and docs

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -136,6 +136,15 @@ struct Cross <: Drawable
     range::UnitRange{Int}
 end
 
+
+"""
+    rectangle = Rectangle(p, x, y)
+A `Drawable` rectangle from position p with length x and width y.
+"""
+
+Rectangle(p::CartesianIndex{2}, x::Int, y::Int) = 
+    Polygon([p,CartesianIndex(p[1],p[2]+y),CartesianIndex(p[1]+x,p[2]+y),CartesianIndex(p[1]+x, p[2])])
+
 """
     img = draw!(img, drawable, color)
     img = draw!(img, drawable)
--- a/src/core.jl
+++ b/src/core.jl
@@ -145,6 +145,12 @@ A `Drawable` rectangle from position p with length x and width y.
 Rectangle(p::CartesianIndex{2}, x::Int, y::Int) = 
     Polygon([p,CartesianIndex(p[1],p[2]+y),CartesianIndex(p[1]+x,p[2]+y),CartesianIndex(p[1]+x, p[2])])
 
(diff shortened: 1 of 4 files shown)


src/core.jl
```diff
@@ -136,6 +136,21 @@ struct Cross <: Drawable
     range::UnitRange{Int}
 end
 
+
+"""
+    rectangle = Rectangle(p, x, y)
```
johnnychen94, line 141: I prefer to not support `Rectangle(p::Point, x, y)` as the meaning of `x` and `y` can be confusing.

Instead, I prefer the following API:

```julia
Rectangle(from::Point, to::Point)
Rectangle(from::CartesianIndex, to::CaretesianIndex)
```

mgautam98: Yes this seems more reasonable. I tried to follow [Matlab's API](https://www.mathworks.com/help/images/ref/images.roi.rectangle.html) but we can do better 😸

src/core.jl
```diff
@@ -136,6 +136,21 @@ struct Cross <: Drawable
...
+
+"""
+    rectangle = Rectangle(p1, p2)
+A `Drawable` rectangle with p1 as top left and p2 as bottom right vertices.
+"""
+
+Rectangle(from::CartesianIndex{2}, to::CartesianIndex{2}) = 
+    Polygon([from,CartesianIndex(from[1],to[2]),to,CartesianIndex(to[1], from[2])])
+
+Rectangle(x1::Int, y1::Int, x2::Int, y2::Int) = Rectangle(CartesianIndex(x1, y1), CartesianIndex(x2, y2))
```
johnnychen94, line 148: I suggest we remove this method as it's very confusing.

```julia
Rectangle(1, 2, 3, 4) # what does this mean?
```

test/paths.jl
```diff
@@ -84,3 +84,14 @@ end
...
+
+
+@testset "Rectangle" begin
+    img = zeros(Gray, 10, 10)
+    expected = copy(img)
+    expected[2, 3:7] .= Gray(1)
+    expected[5, 3:7] .= Gray(1)
+    expected[2:5, 3] .= Gray(1)
+    expected[2:5, 7] .= Gray(1)
+    @test all(expected .== @inferred draw(img, Rectangle((2,3),(5,7)))) == true
```
johnnychen94, line 96: Also need to cover the cases for `CartesianIndex` and `Point` types.

ashwani-rathee:
I can complete this,doesn't seem too difficult...should I give it a go?

mgautam98:
> I can complete this,doesn't seem too difficult...should I give it a go?

Sure! give it a go

PR #51, opened 2021-03-16
add fill and thickness keywords to ellipse and circle

### LINKED ISSUE #50: Discussion on circle fill,circle stroke option
I needed those for example : Detection and measurement,indentification of circular objects in a image
Even though Luxor.jl provides functions for that,would we need it here?Just a simple circular,elliptical drawable is unintuitive when we explain more advanced ideas there
Idea: https://github.com/ashwani-rathee/Demos/blob/main/%F0%9F%8E%88%202.pdf
relevant issues:
https://github.com/JuliaImages/ImageEdgeDetection.jl/issues/21

What I kind of need

@johnnychen94 What do you think?

ashwani-rathee:
I only made changes for circle case which is made through ellipse drawing methods due to the way the loops work...it had me all confused regarding the thickness and how we will calculate it.Though I am not entirely sure if the math in current case is completely right either too but it works just fine for a Initial PR
- Need to add thickness and dims keywords for CircleThree Points,Ellipse(this one got me really confused)

- [x] supposed to fix #50

### DIFF
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -2,8 +2,8 @@ import LinearAlgebra: det
 
 #CirclePointRadius methods
 
-CirclePointRadius(x::Int, y::Int, ρ::T; thickness::Int = 0, fill::Bool = true) where {T<:Real} = CirclePointRadius(Point(x,y), ρ, thickness, fill)
-CirclePointRadius(p::CartesianIndex{2}, ρ::T; thickness::Int = 0, fill::Bool = true) where {T<:Real} = CirclePointRadius(Point(p), ρ, thickness, fill)
+CirclePointRadius(x::Int, y::Int, ρ::T; thickness::Int = 0, fill::Bool = true) where {T<:Real} = CirclePointRadius{T}(Point(x,y), ρ, thickness, fill)
+CirclePointRadius(p::CartesianIndex{2}, ρ::T; thickness::Int = 0, fill::Bool = true) where {T<:Real} = CirclePointRadius{T}(Point(p), ρ, thickness, fill)
 
 draw!(img::AbstractArray{T, 2}, circle::CirclePointRadius, color::T) where {T<:Colorant} = draw!(img, Ellipse(circle), color)
 
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -1,9 +1,8 @@
 import LinearAlgebra: det
 
 #CirclePointRadius methods
-
-CirclePointRadius(x::Int, y::Int, ρ::T) where {T<:Real} = CirclePointRadius(Point(x,y), ρ)
-CirclePointRadius(p::CartesianIndex{2}, ρ::T) where {T<:Real} = CirclePointRadius(Point(p), ρ)
+CirclePointRadius(x::Int, y::Int, ρ::T,thickness::Int,fill::Bool) where {T<:Real} = CirclePointRadius(Point(x,y), ρ,thickness,fill)
+CirclePointRadius(p::CartesianIndex{2}, ρ::T,thickness::Int,fill::Bool) where {T<:Real} = CirclePointRadius(Point(p), ρ,thickness,fill)
 
 draw!(img::AbstractArray{T, 2}, circle::CirclePointRadius, color::T) where {T<:Colorant} = draw!(img, Ellipse(circle), color)
 
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -24,5 +24,5 @@ function draw!(img::AbstractArray{T, 2}, circle::CircleThreePoints, color::T) wh
     ρ = euclidean([x1, y1], R)
     (first(ind[2]) <= R[1] <= last(ind[2]) && first(ind[1]) <= R[2] <= last(ind[1])) || error("Center of circle is out of the bounds of image")
     ρ - 1 <= minimum(abs, [R[1] - first(ind[2]), R[1] - last(ind[2]), R[2] - first(ind[1]), R[2] - last(ind[1])]) || error("Circle is out of the bounds of image : Radius is too large")
-    draw!(img, CirclePointRadius(Point(round(Int,R[1]), round(Int,R[2])), ρ), color)
+    draw!(img, CirclePointRadius(Point(round(Int,R[1]), round(Int,R[2])), ρ, 0, true), color)
 end
--- a/src/circle2d.jl
+++ b/src/circle2d.jl
@@ -1,8 +1,9 @@
 import LinearAlgebra: det
 
 #CirclePointRadius methods
-CirclePointRadius(x::Int, y::Int, ρ::T,thickness::Int,fill::Bool) where {T<:Real} = CirclePointRadius(Point(x,y), ρ,thickness,fill)
-CirclePointRadius(p::CartesianIndex{2}, ρ::T,thickness::Int,fill::Bool) where {T<:Real} = CirclePointRadius(Point(p), ρ,thickness,fill)
+
+CirclePointRadius(x::Int, y::Int, ρ::T; thickness::Int = 0, fill::Bool = true) where {T<:Real} = CirclePointRadius(Point(x,y), ρ, thickness, fill)
+CirclePointRadius(p::CartesianIndex{2}, ρ::T; thickness::Int = 0, fill::Bool = true) where {T<:Real} = CirclePointRadius(Point(p), ρ, thickness, fill)
 
 draw!(img::AbstractArray{T, 2}, circle::CirclePointRadius, color::T) where {T<:Colorant} = draw!(img, Ellipse(circle), color)
 
--- a/src/core.jl
+++ b/src/core.jl
@@ -60,6 +60,8 @@ A `Drawable` circle having center `center` and radius `ρ`
 struct CirclePointRadius{T<:Real} <: Circle
     center::Point
     ρ::T
+    thickness::Int
+	fill::Bool
 end
 
 """
@@ -94,6 +96,8 @@ struct Ellipse{T<:Real, U<:Real} <: Drawable
     center::Point
     ρx::T
     ρy::U
+    thickness::Int
+	fill::Bool
 end
 
 """
--- a/src/core.jl
+++ b/src/core.jl
@@ -53,16 +53,36 @@ struct CircleThreePoints <: Circle
 end
 
 """
-    circle = CirclePointRadius(center, ρ)
+    circle = CirclePointRadius(center, ρ; thickness, fill)
 
 A `Drawable` circle having center `center` and radius `ρ`
+with keyword arguments `thickness` and `fill` for drawing hollow circle
+
+# Arguments
+- `thickness::Integer`: thickness of the circle 
+- `fill::Bool`: argument to determine whether to fill the circle or not
+
+# Examples
+```
+julia> using ImageDraw,TestImages,ImageView, ColorVectorSpace
+julia> img = testimage("lighthouse")
+
+julia> draw!(img, Ellipse(CirclePointRadius(350,200,100))) 
+julia> draw!(img, Ellipse(CirclePointRadius(150,200,100; thickness = 70 , fill = false)))
+
+julia> imshow(img)
+
+```
+
 """
 struct CirclePointRadius{T<:Real} <: Circle
     center::Point
     ρ::T
     thickness::Int
-	fill::Bool
+    fill::Bool
 end
+
+CirclePointRadius(center::Point, ρ::Real; thickness=1, fill=true) = CirclePointRadius(center, ρ, thickness, fill)
 
 """
     ls = LineSegment(p1, p2)
@@ -90,6 +110,7 @@ end
     ellipse = Ellipse(center, ρx, ρy)
 
 A `Drawable` ellipse with center `center` and parameters `ρx` and `ρy`
+with keyword arguments `thickness` and `fill` for drawing hollow ellipse
 
 """
 struct Ellipse{T<:Real, U<:Real} <: Drawable
@@ -97,8 +118,10 @@ struct Ellipse{T<:Real, U<:Real} <: Drawable
     ρx::T
     ρy::U
     thickness::Int
-	fill::Bool
+    fill::Bool
 end
+
+Ellipse(center::Point, ρx::T, ρy::U; thickness::Int =  0, fill::Bool = true) where {T<:Real, U<:Real} = Ellipse(center, ρx, ρy, thickness, fill)
 
 """
     polygon = Polygon([vertex])
--- a/src/core.jl
+++ b/src/core.jl
@@ -53,7 +53,7 @@ struct CircleThreePoints <: Circle
 end
 
 """
-    circle = CirclePointRadius(center, ρ; thickness, fill)
+    circle = CirclePointRadius(center, ρ; thickness = 0, fill = false)
 
 A `Drawable` circle having center `center` and radius `ρ`
 with keyword arguments `thickness` and `fill` for drawing hollow circle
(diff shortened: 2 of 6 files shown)


ashwani-rathee:
That's definitely scary,I might also have to rewrite the tests 😅

johnnychen94:
Looks like you're targeting in the right direction :tada:. I'll check the details when docs and tests are added.

Just one quick question: what's the purpose of `fill` here? I imagine that `fill=false` tells `draw!` to only draw the boundary circle, while `fill=true` to draw both the boundary and the internal pixels of it.

ashwani-rathee:
@johnnychen94 Yeah when fill is false,It tells the program to find the breakpoint of val(to create a certain thickness) which is otherwise 0(means it will reach for the center)

ashwani-rathee:
I am still thinking on how to go about this ,as the functions depend on each other(CircleThreePoints uses CirclePointRadius which again goes to Ellipse)and what's reach of changes we need. (0, true) just means please use fill option/and no breakpoints .
Also need to make them to not need all the parameters,(0,true) became a issue mostly due to that..

ashwani-rathee:
Though,I am still confused on how to make optional arguments out of a structure 😄.will need some guidance on how to go about it..or not go about it

src/circle2d.jl
```diff
@@ -1,9 +1,8 @@
 import LinearAlgebra: det
 
 #CirclePointRadius methods
-
-CirclePointRadius(x::Int, y::Int, ρ::T) where {T<:Real} = CirclePointRadius(Point(x,y), ρ)
-CirclePointRadius(p::CartesianIndex{2}, ρ::T) where {T<:Real} = CirclePointRadius(Point(p), ρ)
+CirclePointRadius(x::Int, y::Int, ρ::T,thickness::Int,fill::Bool) where {T<:Real} = CirclePointRadius(Point(x,y), ρ,thickness,fill)
```
johnnychen94, line 4: Need to avoid format issues like this.

```suggestion
CirclePointRadius(x::Int, y::Int, ρ::T, thickness::Int, fill::Bool) where {T<:Real} = CirclePointRadius(Point(x,y), ρ, thickness, fill)
```

src/core.jl
```diff
@@ -60,6 +60,8 @@ A `Drawable` circle having center `center` and radius `ρ`
 struct CirclePointRadius{T<:Real} <: Circle
     center::Point
     ρ::T
+    thickness::Int
+	fill::Bool
 end
```
johnnychen94, line 65: > I am still confused on how to make optional arguments out of a structure

Adding an outer constructor would be sufficient, for example:

```julia
CirclePointRadius(center::Point, ρ::Real; thickness=1, fill=true) = CirclePointRadius(center, ρ, thickness, fill)
```

src/ellipse2d.jl
```diff
@@ -1,16 +1,23 @@
...
+Ellipse(p::CartesianIndex{2}, ρx::T, ρy::U, thickness::Int ,fill::Bool) where {T<:Real, U<:Real} = Ellipse(Point(p), ρx, ρy,thickness,fill)
+Ellipse(circle::CirclePointRadius) = Ellipse(circle.center, circle.ρ, circle.ρ,circle.thickness,circle.fill)
 
 function draw!(img::AbstractArray{T, 2}, ellipse::Ellipse, color::T) where T<:Colorant
 	ys = Int[]
 	xs = Int[]
+	break_point = 0
+    if ellipse.thickness >= ellipse.ρy || ellipse.thickness >= ellipse.ρx
+			error("Thickness greater than a,b not allowed")
+	end
```
johnnychen94, line 13: and don't mix TAB with 4 SPACES

```suggestion
    if ellipse.thickness >= ellipse.ρy || ellipse.thickness >= ellipse.ρx
        error("Thickness greater than a,b not allowed")
    end
```

ashwani-rathee:
I think the break point calculation is correct ,but I don't know where the glitch(in thickness) is coming from.Maybe it's with the way ellipse points are chosen or maybe from ellipse draw function.This has issues 😅

src/core.jl
```diff
@@ -53,17 +53,37 @@ struct CircleThreePoints <: Circle
...
 
 """
-    circle = CirclePointRadius(center, ρ)
+    circle = CirclePointRadius(center, ρ; thickness, fill)
```
johnnychen94, line 56:
```suggestion
    circle = CirclePointRadius(center, ρ; thickness=1, fill=true)
```

src/core.jl
```diff
@@ -53,17 +53,37 @@ struct CircleThreePoints <: Circle
...
+
+# Examples
+```
+julia> using ImageDraw,TestImages,ImageView, ColorVectorSpace
+julia> img = testimage("lighthouse")
+
+julia> draw!(img, Ellipse(CirclePointRadius(350,200,100))) 
+julia> draw!(img, Ellipse(CirclePointRadius(150,200,100; thickness = 70 , fill = false)))
+
+julia> imshow(img)
```
johnnychen94, line 73:
```suggestion
```julia
using ImageDraw, TestImages, ColorVectorSpace
img = testimage("lighthouse")
draw!(img, Ellipse(CirclePointRadius(350,200,100)))
draw!(img, Ellipse(CirclePointRadius(150,200,100; thickness = 70 , fill = false)))
```

src/core.jl
```diff
@@ -90,16 +110,19 @@ end
...
 struct Ellipse{T<:Real, U<:Real} <: Drawable
     center::Point
     ρx::T
     ρy::U
     thickness::Int
-	fill::Bool
+    fill::Bool
 end
 
+Ellipse(center::Point, ρx::T, ρy::U; thickness::Int =  0, fill::Bool = true) where {T<:Real, U<:Real} = Ellipse(center, ρx, ρy, thickness, fill)
```
johnnychen94, line 124: It's unlikely to reformat the code after being merged so we need to make sure that we do it correctly here. There're still some other similar format issues.

```suggestion
Ellipse(center::Point, ρx::T, ρy::U; thickness::Int = 0, fill::Bool = true) where {T<:Real, U<:Real} = Ellipse(center, ρx, ρy, thickness, fill)
```

src/ellipse2d.jl
```diff
@@ -1,18 +1,18 @@
...
     if ellipse.thickness >= ellipse.ρy || ellipse.thickness >= ellipse.ρx
-			error("Thickness greater than a,b not allowed")
+		error("Thickness greater than a,b not allowed")
 	end
 	if ellipse.fill == false
-        break_point = ((ellipse.ρy - ellipse.thickness)/ ellipse.ρy) ^ 2 + ((ellipse.ρx - ellipse.thickness)/ ellipse.ρx) ^2 
+        break_point = ((ellipse.ρy - ellipse.thickness)/ ellipse.ρy) ^ 2 + ((ellipse.ρx - ellipse.thickness)/ ellipse.ρx) ^ 2 
 	end
 	for i in ellipse.center.y : ellipse.center.y + ellipse.ρy
 		for j in ellipse.center.x : ellipse.center.x + ellipse.ρx
```
johnnychen94, line 18: I still see TABs in this function.

test/circle2d.jl
```diff
@@ -15,19 +15,33 @@ using LinearAlgebra
...
+                               0 0 1 1 1 1 1 1 1 0
+                               0 1 1 1 0 0 0 1 1 1
+                               0 1 1 0 0 0 0 0 1 1
+                               0 1 1 0 0 0 0 0 1 1
+                               0 1 1 0 0 0 0 0 1 1
+                               0 1 1 1 0 0 0 1 1 1
+                               0 0 1 1 1 1 1 1 1 0
+                               0 0 0 1 1 1 1 1 0 0 ]
+
+        res = @inferred draw(img, CirclePointRadius(Point(6, 6), 5, 3, false))
```
johnnychen94, line 43: We also need tests to make sure

```julia
@test CirclePointRadius(Point(6, 6), 5, 3, false) == 
CirclePointRadius(Point(6, 6), 5, thickness=3, fill=false)
```

and some invalid `thickness` input to make sure it errors properly, e.g.,:

```julia
err = ArgumentError("Thickness 7 should be smaller than 5.")
@test_throws CirclePointRadius(Point(6, 6), 5, thickness=7, fill=false)
```

src/ellipse2d.jl
```diff
@@ -1,18 +1,18 @@
...
+Ellipse(circle::CirclePointRadius) = Ellipse(circle.center, circle.ρ, circle.ρ, circle.thickness, circle.fill)
 
 function draw!(img::AbstractArray{T, 2}, ellipse::Ellipse, color::T) where T<:Colorant
 	ys = Int[]
 	xs = Int[]
 	break_point = 0
     if ellipse.thickness >= ellipse.ρy || ellipse.thickness >= ellipse.ρx
-			error("Thickness greater than a,b not allowed")
+		error("Thickness greater than a,b not allowed")
 	end
```
johnnychen94: We need to early raise errors whenever we can, so the "correct" place to do these argument checks are in the inner constructor:

```julia
struct Ellipse{T<:Real, U<:Real} <: Drawable
    center::Point
    ρx::T
    ρy::U
    thickness::Int
    fill::Bool
    function Ellipse{T, U}(center, ρx::T, ρy::U, thickness::Int, fill::Bool) where {T<:Real, U<:Real}
        thickness >= min(ρx, ρy) && throw(ArgumentError("Thickness $thickness should be smaller than $(min(ρx, ρy))."))
        new{T, U}(center, ρx, ρy, thickness, fill)
    end
end
```

src/core.jl
```diff
@@ -53,17 +53,37 @@ struct CircleThreePoints <: Circle
...
 
 """
-    circle = CirclePointRadius(center, ρ)
+    circle = CirclePointRadius(center, ρ; thickness, fill)
 
 A `Drawable` circle having center `center` and radius `ρ`
+with keyword arguments `thickness` and `fill` for drawing hollow circle
+
+# Arguments
+- `thickness::Integer`: thickness of the circle
```
johnnychen94, line 62: If you declare `thickness::Integer` is supported, then you need to add such tests for them, e.g., to make sure that `thickness=UInt8(1)` works.

johnnychen94: Also, the argument explanation should exist in docs of other structs, too. You won't expect the user of `Ellipse` to query the meaning of `thickness` by checking `?CirclePointRadius`.

ashwani-rathee: Sure,I'll fix all those issues..can you check the way thickness is to be implemented in a ellipse,since our circle draw is dependent on ellipse(otherwise it would have quite straightforward) and the thickness of elliptic curve is a difficult topic with parallel curves to be drawn

johnnychen94: I'm quite busy learning halide and the C++ binding toolchain recently 😢  I'll check it this weekend if you haven't figured it out then.

johnnychen94:
> I think the break point calculation is correct ,but I don't know where the glitch(in thickness) is coming from.Maybe it's with the way ellipse points are chosen or maybe from ellipse draw function.

The result looks good to me. Can you explain what the issue is here? A reproducible code would be very much helpful for me to understand your question.
It would also be helpful to look at the code if you can address the unsolved comments first.

ashwani-rathee:
@johnnychen94 sure I'll update it with fixes,just was caught up in something rather urgent :)

src/core.jl
ashwani-rathee: Other than this one,I have added fixes for others .Inner and outer constructor got me quite confused especially with keyword arguments since the some test cases were able to see the inner constructor or maybe I missed something 😓 

> If you declare `thickness::Integer` is supported, then you need to add such tests for them, e.g., to make sure that `thickness=UInt8(1)` works.

ashwani-rathee:
I have tried to pull changes but unable to for some reason and when I try to directly resolve ,it takes a long time to merge

src/ellipse2d.jl
```diff
@@ -1,16 +1,20 @@
...
+Ellipse(p::CartesianIndex{2}, ρx::T, ρy::U; thickness::Int = 0, fill::Bool = true) where {T<:Real, U<:Real} = Ellipse{T, U}(Point(p), ρx, ρy, thickness, fill)
+Ellipse(circle::CirclePointRadius) = Ellipse(circle.center, circle.ρ, circle.ρ;  thickness = circle.thickness, fill = circle.fill)
 
 function draw!(img::AbstractArray{T, 2}, ellipse::Ellipse, color::T) where T<:Colorant
 	ys = Int[]
 	xs = Int[]
+	break_point = 0
+	if ellipse.fill == false
+        break_point = ((ellipse.ρy - ellipse.thickness)/ ellipse.ρy) ^ 2 + ((ellipse.ρx - ellipse.thickness)/ ellipse.ρx) ^ 2 
+	end
```
johnnychen94, line 13: The format here is suspicious; I still see TAB and spaces.

src/core.jl
johnnychen94:
```suggestion
- `thickness::Int`: thickness of the circle
```

src/core.jl
```diff
@@ -86,18 +110,42 @@ struct Path <: Drawable
...
 
 """
-    ellipse = Ellipse(center, ρx, ρy)
+    ellipse = Ellipse(center, ρx, ρy; thickness = 0,fill = false)
 
 A `Drawable` ellipse with center `center` and parameters `ρx` and `ρy`
+with keyword arguments `thickness` and `fill` for drawing hollow ellipse
 
+# Arguments
+- `thickness::Integer`: thickness of the ellipse
```
johnnychen94, line 119:
```suggestion
- `thickness::Int`: thickness of the circle
```

test/circle2d.jl
```diff
@@ -28,6 +28,29 @@ using LinearAlgebra
         expected = map(i->(i==1 ? Gray{N0f8}(0.5) : Gray{N0f8}(0)), expected)
         @test all(expected .== res) == true
 
+        img = zeros(Gray{N0f8}, 10, 10)
```
johnnychen94, line 31: For each type, we need one more test for the default value, e.g., 

```julia
draw(img, CirclePointRadius(Point(6, 6), 5; thickness = 0, fill = true)) == draw(img, CirclePointRadius(Point(6, 6), 5))
```

This looks tedious but it is an important test case.

johnnychen94 (approved):
It would be nice to also have thickness and fill keyword supported for other shapes. But as far as you've worked in this PR, it is great already.

src/core.jl
ashwani-rathee: Only test that needs to be resolved is the UInt8 one,I think we can work on those keywords for other shapes after Polygon filling work is done....that would give us more perspective and how to best implement those keywords :)

ashwani-rathee: WARNING: Method definition (::Type{ImageDraw.CirclePointRadius{T} where T<:Real})(ImageDraw.Point, T<:Real) where {T<:Real} in module ImageDraw at /home/ashwani/Desktop/ImageDraw.jl/src/core.jl:87 overwritten at /home/ashwani/Desktop/ImageDraw.jl/src/core.jl:88.
  ** incremental compilation may be fatally broken for this module **
  Also this issue needs to be fixed before merging

johnnychen94: Ah, okay. This means that we also need test cases for `CartesianIndex`. I only see `Point` version used in the tests.

ashwani-rathee: @johnnychen94 Can you point out which cases I need to check for cartesian index?
```julia
        res = @inferred draw(img, CirclePointRadius(Point(5,5), 5), Gray{N0f8}(0.5))
        expected = map(i->(i==1 ? Gray{N0f8}(0.5) : Gray{N0f8}(0)), expected)
        @test all(expected .== res) == true
```
I am assuming this one

PR #52, opened 2021-03-24
Add RectanglePoints

ashwani-rathee:
Based off ideas from #41 ,some additional things
- Tested on local system,test pass
- Haven't checked the documentation pass,but hopefully will pass
- Docstring added
- Implemented using Polygon functionality

Future work: 
- fill option
- square box using this
- Dotted,solid,thickness needs to be worked out

I know it can be more concise :)

closes #41

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -35,22 +35,19 @@ When displayed, these three images look like this:
 <p>
 ```
 
-```
 
 Drawing a Rectangle.
 
 ```@example usage
-using TestImages, ImageDraw, ImageCore
+using TestImages, ImageDraw, ImageCore, ImageShow
 using FileIO # hide
 img = testimage("lighthouse")
-img_example = draw!(img, Polygon(RectanglePoints(Point(1, 1), Point(5, 9))), RGB{N0f8}(1))
-save("images/lighthouse_rectangle.png", img_example) # hide
+
+img_example_stage1 = draw!(img, Polygon(RectanglePoints(Point(10, 10), Point(100, 100))), RGB{N0f8}(1))
+img_example_stage2 = draw!(img_example_stage1, Polygon(RectanglePoints(CartesianIndex(110, 10), CartesianIndex(200, 200))), RGB{N0f8}(1))
+img_example_stage3 = draw!(img_example_stage2, Polygon(RectanglePoints(220, 10, 300, 300)), RGB{N0f8}(1))
 ```
 
-```@raw html
-<img src="images/lighthouse_rectangle.png" width="512px" alt="lighthouse_rectangle" />
-<p>
-```
 
 Drawing a Cross.
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -6,7 +6,7 @@ A drawing package for JuliaImages
 Depth = 3
 ```
 
-## introduction
+## Introduction
 
 ImageDraw supports basic drawing on Images. You can draw points, lines, circles, ellipse and paths.  
 
@@ -32,6 +32,23 @@ save("images/lighthouse_circle.png", img) # hide
 When displayed, these three images look like this:
 ```@raw html
 <img src="images/lighthouse_circle.png" width="512px" alt="edge detection demo 1 image" />
+<p>
+```
+
+```
+
+Drawing a Rectangle.
+
+```@example usage
+using TestImages, ImageDraw, ImageCore
+using FileIO # hide
+img = testimage("lighthouse")
+img_example = draw!(img, Polygon(RectanglePoints(Point(1, 1), Point(5, 9))), RGB{N0f8}(1))
+save("images/lighthouse_rectangle.png", img_example) # hide
+```
+
+```@raw html
+<img src="images/lighthouse_rectangle.png" width="512px" alt="lighthouse_rectangle" />
 <p>
 ```
 
--- a/src/core.jl
+++ b/src/core.jl
@@ -110,10 +110,26 @@ struct Polygon <: Drawable
 end
 
 """
-    rectangle = RectanglePoints(p1,p2)
+    rectangle = RectanglePoints(p1, p2)
+    rectangle = RectanglePoints(x1, y1, x2, y2)
 
 A `Drawable` rectangle i.e. a closed path where parameters `p1` and `p2` 
 are diagonally opposite vertices of the rectangle.
+
+Parameters `p1` and `p2` can be passed in Point struct format or CartesianIndex{} format as shown in example below.
+
+# Examples
+```julia
+using TestImages, ImageDraw, ImageCore
+img = testimage("lighthouse")
+
+# rectangles drawn on img using different methods of passing parameters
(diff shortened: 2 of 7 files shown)


docs/src/index.md
```diff
@@ -35,6 +35,23 @@ When displayed, these three images look like this:
 <p>
 ```
 
+```
```
johnnychen94, line 38: This seems to be redundant to me.

docs/src/index.md
```diff
@@ -35,6 +35,23 @@ When displayed, these three images look like this:
...
+
+Drawing a Rectangle.
+
+```@example usage
+using TestImages, ImageDraw, ImageCore
+using FileIO # hide
+img = testimage("lighthouse")
+img_example = draw!(img, Polygon(RectanglePoints(Point(1, 1), Point(5, 9))), RGB{N0f8}(1))
+save("images/lighthouse_rectangle.png", img_example) # hide
```
johnnychen94, line 47: IIUC the `example` block automatically inserts the result of the last expression so you don't need to save it. 

```suggestion
using TestImages, ImageDraw, ImageCore, ImageShow
img = testimage("lighthouse")
img_example = draw!(img, Polygon(RectanglePoints(Point(1, 1), Point(5, 9))), RGB{N0f8}(1))
```

`ImageShow` is required to reduce the displayed image size because otherwise it will save images using SVG format, which is crazy for normal image types.

src/core.jl
```diff
@@ -108,6 +109,17 @@ struct Polygon <: Drawable
     vertices::Vector{Point}
 end
 
+"""
+    rectangle = RectanglePoints(p1,p2)
+
+A `Drawable` rectangle i.e. a closed path where parameters `p1` and `p2` 
+are diagonally opposite vertices of the rectangle.
+"""
```
johnnychen94, line 117: Would also be nice to have some examples here.

```suggestion
"""
    rectangle = RectanglePoints(p1, p2)
    rectangle = RectanglePoints(x1, y1, x2, y2)

A `Drawable` rectangle i.e. a closed path where parameters `p1` and `p2` 
are diagonally opposite vertices of the rectangle.
"""
```

johnnychen94 (commented):
Very nicely done!

ashwani-rathee:
Updated with docs and docstring improvments

johnnychen94 (approved):

ashwani-rathee:
I am thinking about Rectangle Fill method now,I tried several things  and need to know which method works best
Since Rectange fits right into how we index and slice arrays,I tried to do that by broadcasting values,using fill!,though single value change works but not all of them or atleast the result doesn't propagate back to original image we want.
I am sure,I am missing something
```julia
using Images
img =testimage("lighthouse")
img[100,200] =RGB{N0f8(1) #this works
img[100:100,200:200].=RGB{N0f8}(1) #this doesn't,and neither does RGB{N0f8}(1,1,1)

#this also doesn't work atleast with what little I know..
#.also tried @view img[:,:] and copy(img) but update on image don't propagate for me
julia> img
512×768 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}:
julia> img_2=img[100:100,200:200]
1×1 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}:
 RGB{N0f8}(1.0,1.0,1.0)
julia> fill!(img_2,1)
1×1 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}:
 RGB{N0f8}(1.0,1.0,1.0)
julia> fill!(img_2,RGB(1,1,1))
1×1 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}:
 RGB{N0f8}(1.0,1.0,1.0)
julia> imshow(img)
Dict{String,Any} with 4 entries:
  "gui"         => Dict{String,Any}("window"=>GtkWindowLeaf(name="", parent, width-request=-1, height-request=-1, visible=TRUE, s…
  "roi"         => Dict{String,Any}("redraw"=>549: "map(clim-mapped image, input-280)" = nothing Nothing ,"zoomregion"=>500: "inp…
  "annotations" => 502: "input-280" = Dict{UInt64,Any}() Dict{UInt64,Any} 
  "clim"        => 501: "CLim" = CLim{RGB{Float64}}(RGB{Float64}(0.0,0.0,0.0), RGB{Float64}(1.0,1.0,1.0)) CLim{RGB{Float64}} 

#did explore channelview,rawview due to this and ImageCore.jl is not directly in front of documentation even
# though it has good relevant content regarding this
```
Ends the rant of what little I know,there has to be a way of doing this that I don't know yet😅

Now,the solution that works now
```julia
RectanglePoints(x1::Int, y1::Int, x2::Int, y2::Int, fill::Int)  = RectanglePoints(Point(x1, y1), Point(x2, y2), fill)
function draw!(img::AbstractArray{T, 2}, rectangle::RectanglePoints, color::T, fill = false) where {T<:Colorant}
    arr=[]
    for i in rectangle.p1.y:rectangle.p2.y
        push!(arr,CartesianIndex(rectangle.p1.x,i))
        push!(arr,CartesianIndex(rectangle.p2.x,i))
    end
    for i in 1:2:length(arr)-1
            draw!(img, LineSegment(arr[i], arr[i+1]), color)
    end
    img
end
``` 
Result:

Explanation:


This all will not be needed after Polygon fill is implemented,Boundary Fill or Flood fill algorithm seems like the way to go if we directly want to jump to that...

ashwani-rathee:
I think this one works,it was fun exploring but we can jump directly to flood fill algorithm in both circle and polygon methods

johnnychen94:
For a rectangle, you can fill a view of the image with a single colorant:

```julia
fill!(view(img, fill_axes), color))
```

Note that when you modify the view, it is `img` that gets modified.

ashwani-rathee:
> I might not understand your question well, so curious to ask, why do you say `img[100:100,200:200] .= RGB{N0f8}(1)` doesn't work?
Actually it works for its specfic part,like it does change values to white,but when I do imshow(img),I don't see any changes on the complete image

johnnychen94:
Could you provide a minimal reproducible example to explain what "it doesn't work" means?

```julia
julia> using Images, TestImages

julia> img = testimage("lighthouse");

julia> img[1:1, 1:1] .= RGB{N0f8}(1)
1×1 view(::Array{RGB{N0f8},2}, 1:1, 1:1) with eltype RGB{N0f8}:
 RGB{N0f8}(1.0,1.0,1.0)

julia> img[1, 1] == RGB{N0f8}(1)
true
```

I see `true` here so `img` is changed.

ashwani-rathee:
> I see `true` here so `img` is changed.
Everything works just fine,I messed up the indexing of the change I needed.. 😅

johnnychen94:
Do you plan to add `fill` in this PR? Or I'll just merge this one.

ashwani-rathee:
we can skip it for now,since I'm working on polygon filling...when the PR for that will come,I'll update these and lot other keywords

johnnychen94:
It seems that there'll be many PRs in so I'm not tagging a new release here. If you need a new release you could just ask me.

Issue #53, opened 2021-03-25
Polygon/Any closed path filling algorithms


ashwani-rathee:
Since this project is still WIP,these filling algorithms commonly used in computer graphics would be quite useful addition
- Boundary Fill Algorithm
 - 4- pixel method
 - 8- pixel method
- Flood fill algorithm
- Scan-line polygon filling

References:
- https://www.tutorialspoint.com/computer_graphics/polygon_filling_algorithm.htm
- Bounday Fill method :https://www.geeksforgeeks.org/boundary-fill-algorithm/
- Flood fill method : https://www.geeksforgeeks.org/flood-fill-algorithm-implement-fill-paint/
- Scan-line polygon filling: https://www.geeksforgeeks.org/scan-line-polygon-filling-using-opengl-c/

ashwani-rathee:
Worked out the Boundary fill algorithm
test2.png: https://i.imgur.com/ImGWeJm.png
Made with RectanglePoints and updated CirclePointRadius :)
![](https://i.imgur.com/ImGWeJm.png)
MRE:
```julia
julia>  using Images,ImageDraw,ColorVectorSpace
julia> function boundaryfill(x::Int, y::Int, fill_color::T, boundary_color::T) where T<:Colorant
               if (img[y, x] != boundary_color && img[y, x] !=fill_color)
                       if checkbounds(Bool, img, y, x) img[y, x] = fill_color end
                        boundaryfill(x + 1, y, fill_color, boundary_color)
                        boundaryfill(x, y + 1, fill_color, boundary_color)
                        boundaryfill(x - 1, y, fill_color, boundary_color)
                        boundaryfill(x, y - 1, fill_color, boundary_color)
               end        
               img
       end
boundaryfill (generic function with 1 method)
julia> img=load("test2.png")
512×768 Array{RGB{N0f8},2} with eltype RGB{Normed{UInt8,8}}
julia> img_new=boundaryfill(120,120,RGB(1),RGB(1))
julia> imshow(img)
julia> @time img_new=boundaryfill(120,120,RGB(1),RGB(1))  #for circle or any other polygon for that matter
0.013602 seconds (338.09 k allocations: 5.159 MiB)
julia> @time img_new=boundaryfill(20,20,RGB(1),RGB(1))   #for square,rectangle
0.003581 seconds (87.13 k allocations: 1.330 MiB)
```
Result:
![](https://i.imgur.com/UN8IGVy.png)

[...]

PR #57, opened 2021-04-03
Polygon filling API - Boundary Fill

ashwani-rathee:
Initial setup of the Polygon filling API
- Implementation of Boundary Fill
Linked issue :https://github.com/JuliaImages/ImageDraw.jl/issues/53
Demonstration : https://github.com/ashwani-rathee/Demonstrations/blob/main/boundary-fill-api.jl
For some reason,the tests in poygon2d.jl isn't able to access boundaryfilll ,I am not sure on why as of now


src/polygonfill2d.jl
```diff
@@ -0,0 +1,57 @@
...
+#     boundarycolor::T
+#     function boundaryfill(x::Int,y::Int,fill_color::T,boundarycolor::T) where {T<:Colorant}
+#         println("Test 1")
+#         new{T}(x,y,fill_color,boundarycolor)
+#     end
+# end
+
+# boundaryfill(;x::Int=0, y::Int=0, fill_color::Colorant=RGB(1), boundarycolor::Colorant=RGB(1)) = boundaryfill(x, y, fill_color, boundarycolor)
+
+function boundaryfill(res::AbstractArray{T,2}, x::Int, y::Int, fill_color::T, boundary_color::T) where T<:Colorant
```
johnnychen94, line 20: You're mixing the usage of function and structure here.

src/polygonfill2d.jl
```diff
@@ -0,0 +1,57 @@
...
+	for i in 1:length(verts)-1
+		draw!(res, LineSegment(verts[i], verts[i+1]), fill_color)
+    end	
+	boundaryfill(res,f.x, f.y, f.fill_color, f.boundarycolor)
+   res
+end
+
+
+
+function draw(img::AbstractArray{T,2}, verts::Vector{CartesianIndex{2}}, f::AbstractPolyFillAlgorithm) where T <: Colorant
```
johnnychen94, line 45: Should also add the `draw!` method.

src/core.jl
```diff
@@ -109,6 +111,24 @@ struct Polygon <: Drawable
     vertices::Vector{Point}
 end
 
+"""
+    boundaryfill(x::Int, y::Int, fill_color::T, boundarycolor::T)
+
+
+"""
+struct boundaryfill{T<:Colorant} <: AbstractPolyFillAlgorithm
```
johnnychen94, line 119: struct and module should normally use the CapWords convention.

So it should be `BoundaryFill` here.

src/polygonfill2d.jl
ashwani-rathee: Somehow API's draw! method breaks draw! method of ImageDraw.LineSegment,even though they are pretty different from each other. Also I am not sure why I am not able to access BoundaryFill in tests.

src/ImageDraw.jl
```diff
@@ -51,4 +52,7 @@ export
 	#Cross
 	Cross
 
+	#Polygon fill
+	BoundaryFill
```
johnnychen94, line 56: Without the comma, `BoundaryFill` is not exported.

```suggestion
      Cross,

      #Polygon fill
      BoundaryFill
```

src/polygonfill2d.jl
johnnychen94: > Somehow API's draw! method breaks draw! method of ImageDraw.LineSegment

Could you describe with more information what "break" means here and how you implemented the `draw!` method? It's a bit hard to imagine what the reason it could be without the actual code and the error message.

src/core.jl
```diff
@@ -109,6 +118,61 @@ struct Polygon <: Drawable
     vertices::Vector{Point}
 end
 
+"""
+    BoundaryFill <: AbstractPolyFillAlgorithm
+    BoundaryFill(x::Int, y::Int, fill_color::T, boundarycolor::T)
+
+    draw(img, verts, BoundaryFill(x, y, fill_color, boundary_color); connectverts)
+    draw!(img, verts, BoundaryFill(x, y, fill_color, boundary_color); connectverts)
```
johnnychen94, line 126: I feel `connectverts` should be documented as well... Generally, we use underscore to split the words, i.e., `connect_verts`. However, I don't know if `verts` are something that's widely used so have no idea if this is a good name.

test/polygon2d.jl
```diff
@@ -0,0 +1,12 @@
+@testset "Polygon2D" begin
+    img = zeros(RGB, 7, 7)
+    expected = copy(img)
+    expected[2:6, 2:6] .= RGB{N0f8}(1)
+    verts = [CartesianIndex(2, 2), CartesianIndex(2, 6), CartesianIndex(6, 6), CartesianIndex(6, 2), CartesianIndex(2,2)]
+    
+    res = @inferred draw(img, verts, BoundaryFill(x=4, y=4, fill_color=RGB(1), boundary_color=RGB(1)); connectverts=true)
```
johnnychen94, line 7: Should also check `connectverts=false`.

src/core.jl
```diff
@@ -109,6 +118,61 @@ struct Polygon <: Drawable
...
+
+# Output
+
+Return the boundary filled image inside the vertices(provided in order). 
+
+# Arguments
+
+## `x` && `y`
+
+Boundary fill is a seeded polygon filling algorithm.So we need to provide the seed (x,y) inside the image.
```
johnnychen94, line 138: I still can't figure out what `x` and `y` are...

src/polygonfill2d.jl
```diff
@@ -0,0 +1,85 @@
...
+"""
+function draw!(img::AbstractArray{T,2}, verts::Vector{CartesianIndex{2}}, f::AbstractPolyFillAlgorithm;connectverts::Bool = true) where T <: Colorant
+    res = copy(img)
+    if (connectverts == true)
+        for i in 1:length(verts) - 1
+            draw!(res, LineSegment(verts[i], verts[i + 1]), f.fill_color)
+        end	
+    end
+    f(res, verts, f.x, f.y, f.fill_color, f.boundary_color)
+end
```
johnnychen94, line 50: A method with `!` suffix is a Julia **Convention** indicating that this function is an in-place function, hence `draw!` should not allocate new memory with `copy(img)`. Here we should directly operate on `img` and users should and will expect that `img` will be changed after `draw!`.

src/polygonfill2d.jl
```diff
@@ -0,0 +1,85 @@
...
+```
+"""
+function draw(img::AbstractArray{T,2}, verts::Vector{CartesianIndex{2}}, f::AbstractPolyFillAlgorithm; connectverts::Bool=true) where {T <: Colorant}
+    res = copy(img)
+    if (connectverts == true)
+        for i in 1:length(verts) - 1
+            draw!(res, LineSegment(verts[i], verts[i + 1]), f.boundary_color)
+        end	
+    end
+    f(res, verts, f.x, f.y, f.fill_color, f.boundary_color)
```
johnnychen94, line 84: For better readbility, please don't copy codes around:

```suggestion
    draw!(copy(img), verts, f; connectverts=connectverts))
```

src/polygonfill2d.jl
```diff
@@ -0,0 +1,85 @@
...
+
+Draw on `img` using algorithm `f`.
+
+# Output
+
+When `img` is specified, a copy of `img` is made and changes are made on it and returned.
+
+# Example
+
+Just simply pass an algorithm with parameters,with image and vertices of polygon
```
johnnychen94, line 64: Well-formatted English is necessary; space is NEEDED after punctuations:

```suggestion
Just simply pass an algorithm with parameters, with image and vertices of polygon
```

I didn't read all your docs, so there might be other similar issues, hope you can avoid this issue next time.

src/core.jl
```diff
@@ -109,6 +118,61 @@ struct Polygon <: Drawable
...
+
+struct BoundaryFill{T<:Colorant} <: AbstractPolyFillAlgorithm
+	x::Int
+	y::Int
+	fill_color::T
+    boundary_color::T
+    function BoundaryFill(x::Int, y::Int, fill_color::T, boundary_color::T) where {T<:Colorant}
+        new{T}(x, y, fill_color, boundary_color)
+    end
+end
```
johnnychen94, line 171: I really don't want to mention this, please do not mix the usage of SPACEs and TABs.

src/core.jl
```diff
@@ -109,6 +118,61 @@ struct Polygon <: Drawable
...
+draw(img, verts, BoundaryFill(x=4, y=4, fill_color=RGB(1), boundary_color=RGB(1)); connectverts=true)
+
+```
+"""
+
+struct BoundaryFill{T<:Colorant} <: AbstractPolyFillAlgorithm
+	x::Int
+	y::Int
+	fill_color::T
+    boundary_color::T
```
johnnychen94, line 167: `fill_value` and `boundary_value` are better names as we accept plain values ( e.g., `0`) instead of Colorant.

src/core.jl
johnnychen94, line 126:
```suggestion
    draw(img, verts, alg::BoundaryFill; connectverts)
    draw!(img, verts, alg::BoundaryFill; connectverts)
````

**Review by johnnychen94 — COMMENTED** [review]
Don't rush, be patient, do less copy. Be careful about your format. Try to read, analyze and find out what the real issue is before asking for help.

I'm asking this because I hope I can spend more time reviewing and discussing more PR-related issues instead of those trivial things that you could have fixed by yourself.

File: src/core.jl (hunk 11 above)
Comment (ashwani-rathee) [reply src/core.jl:171]: Okay, I read the Julia's code style suggestions today. I was confused before if we have to use tab or space since both TABs and SPACEs have been used in ImageDraw.jl.
Indentation Level => 4 spaces everywhere

File: src/core.jl (hunk 5 above)
Comment (ashwani-rathee) [reply src/core.jl:126]: I think `connect_vertices` would be more clear but looks wordy. And it's an important parameter because boundary fill won't know when to stop without connected vertices, flood fill and scan fill don't really care about connected vertices.

Comment (johnnychen94) [reply src/core.jl:126]: > I think `connect_vertices` would be more clear but looks wordy. And it's an important parameter because boundary fill

This looks like a good name to me. Will it be too often that the user needs to change the default behavior, i.e., `connect_vertices=false`? If `false` is rarely used, then I think it's totally fine to use some long but clear words.

Does `connect_vertices` mean the fill range is `closed` region? If so, maybe we can use `closed` as a keyword.

File: src/core.jl (hunk 11 above)
Comment (johnnychen94) [reply src/core.jl:171]: Yes, it's a bit annoying when both SPACEs and TABs are around, so whenever these mixture usages are detected, feel free to change them to 4 SPACEs.

Just that don't change unrelated codes (even if you find they're using the wrong indentation), because that would pollute the git history with no meaningful changes. This is exactly why we want to make sure that we format the codes well; we don't have many chances to correct them after committing to the git history.

File: src/core.jl (hunk 5 above)
Comment (ashwani-rathee) [reply src/core.jl:126]: `connect_vertices` won't be changed much and yes it means the fill range will be closed region. I think `closed` would be good too. Flood fill implementation would make things more clear for me, since it approaches things differently and then fitting them all in a single API would be interesting.

Comment (johnnychen94) [reply src/core.jl:126]: `closed` looks like a better word for `draw` because it's declarative. Users might not care about implementation details(e.g., how to make it a closed range with `connect_vertices`). If other algorithms have their own way to make a closed region, the same keyword `closed` can be reused then.

File: src/core.jl [hunk 13]
```diff
@@ -109,6 +118,66 @@ struct Polygon <: Drawable
...
+// To-do checking if verts and seed are inside the image or note
+
+## `fill_value`
+
+The pixels inside the boundary of vertices of polygon are filled with this colorant parameter.
+
+## `boundary_value`
+
+The pixels to define the boundary using verts is filled with this colorant parameter. 
+Without this parameter, the whole picture will be filled with the `fill_value`
```
johnnychen94, line 149: Is it? Add a test to make sure this is correct.

```diff
- BoundaryFill(;x::Int = 0, y::Int = 0, fill_value::Colorant = RGB(1), boundary_value::Colorant = RGB(1)) = 
+ BoundaryFill(;x::Int = 0, y::Int = 0, fill_value::Colorant = RGB(1), boundary_value::Colorant = fill_value) =
  BoundaryFill(x, y, fill_value, boundary_value)
```

src/core.jl
```diff
@@ -109,6 +118,66 @@ struct Polygon <: Drawable
...
+
+Return the boundary filled image inside the vertices(provided in order). 
+
+# Arguments
+
+## `x` && `y`
+
+Boundary fill is a seeded polygon filling algorithm. 
+So we need to provide the seed point (x,y) inside the image from where the algorithm can start its function.
+// To-do checking if verts and seed are inside the image or note
```
johnnychen94, line 140: Does this mean that you still need extra test cases to check this?

src/core.jl
```diff
@@ -109,6 +118,66 @@ struct Polygon <: Drawable
...
+    x::Int
+    y::Int
+    fill_value::T
+    boundary_value::T
+    function BoundaryFill(x::Int, y::Int, fill_value::T, boundary_value::T) where {T <: Colorant}
+        new{T}(x, y, fill_value, boundary_value)
+    end
+end
+
+BoundaryFill(;x::Int = 0, y::Int = 0, fill_value::Colorant = RGB(1), boundary_value::Colorant = RGB(1)) = BoundaryFill(x, y, fill_value, boundary_value)
```
johnnychen94, line 178: Do we need to support `CartesianIndex(x, y)` or `Point(x, y)` here?

johnnychen94, line 178: I don't see a need to make `x` and `y` keyword parameters. Second to my previous comment, if we plan to support `CartesianIndex`, how would you design the code? (`BoundaryFill(; x::Int=0, y::Int=0, point=CartesianIndex(0, 0), ...)` would just be a mess...)

```suggestion
BoundaryFill(x::Int = 0, y::Int = 0; fill_value::Colorant = RGB(1), boundary_value::Colorant = RGB(1))
```

src/polygonfill2d.jl
```diff
@@ -0,0 +1,98 @@
...
+    verts::Vector{CartesianIndex{2}},
+    f::AbstractPolyFillAlgorithm;
+    closed::Bool = false,
+    ) where {T<:Colorant}
+    if (closed == true)
+        for i = 1:length(verts)-1
+            draw!(img, LineSegment(verts[i], verts[i+1]), f.fill_value)
+        end
+    end
+    f(img, verts, f.x, f.y, f.fill_value, f.boundary_value)
```
johnnychen94, line 62: You don't need to pass these to `f()` because it's already known to `f`.

```suggestion
    f(img, verts)
```

src/polygonfill2d.jl
```diff
@@ -0,0 +1,98 @@
...
+```
+"""
+function draw(
+    img::AbstractArray{T,2},
+    verts::Vector{CartesianIndex{2}},
+    f::AbstractPolyFillAlgorithm;
+    closed::Bool = false,
+    ) where {T<:Colorant}
+    draw!(copy(img), verts, f; closed = closed)
+end
```
johnnychen94, line 98: This way we don't mess up with default parameters. And potentially support N-dimensional array, although we may never use it.

```suggestion
function draw(
    img::AbstractArray{<:Colorant,N},
    verts::Vector{CartesianIndex{N}},
    f::AbstractPolyFillAlgorithm;
    kwargs...
    ) where N
    draw!(copy(img), verts, f; kwargs...)
end
```

src/polygonfill2d.jl
```diff
@@ -0,0 +1,98 @@
...
+    boundary_value::T,
+    ) where {T<:Colorant}
+    if checkbounds(Bool, res, y, x)
+        if (res[y, x] != f.boundary_value && res[y, x] != f.fill_value)
+            res[y, x] = f.fill_value
+            f(res, verts, x + 1, y, fill_value, boundary_value)
+            f(res, verts, x, y + 1, fill_value, boundary_value)
+            f(res, verts, x - 1, y, fill_value, boundary_value)
+            f(res, verts, x, y - 1, fill_value, boundary_value)
+        end
```
johnnychen94, line 22: > You don't need to pass these to f() because it's already known to f.

What's the difference between `f.fill_value` and `fill_value`?

johnnychen94 (commented):
The test failure on Julia 1.0 should be fixed by #59 It's because pkg resolving in old Julia versions are not smart enough in some cases.

Mostly it looks good to me now. Hopefully, this is the last round review.

If you're not confident about the implementation, you can always add more tests to make sure of it.

src/core.jl
ashwani-rathee: I am not sure if it's required, because vertices and seed are supposed to be inside to make logical sense. Since we want to draw inside the image and that seed need to be inside the closed polygon formed when vertices(these are to be provided in order) are given..

johnnychen94: > I am not sure if it's required, because vertices and seed are supposed to be inside to make logical sense

The `boundscheck` looks good to me. As a regression test, it can be useful to detect if our logic/implementation is correct.

johnnychen94: Would be fine if writing such a test being too difficult, since we only have very limited image morphology support. (Making it hard to detect the region).

src/core.jl
ashwani-rathee: Sure

src/core.jl
ashwani-rathee: That would be a good addition I think too 👍🏻

src/polygonfill2d.jl
ashwani-rathee: They aren't different, (f.x , f.y ) and (x,y) are different as former is initial value and later pair keeps changing.

johnnychen94:
@ashwani-rathee This PR a little bit long and old. Would you mind summarizing a little bit so that we can move forward quicker?

ashwani-rathee:
@johnnychen94 In the latest improvements( 2 days back before that were rebase merge changes), 12 tests were added and methods for cartesian index, point were added and bounds check was added for vertices and seed point.

src/core.jl
```diff
@@ -223,7 +223,9 @@ struct BoundaryFill{T<:Colorant} <: AbstractPolyFillAlgorithm
     end
 end
 
-BoundaryFill(;x::Int = 0, y::Int = 0, fill_value::Colorant = RGB(1), boundary_value::Colorant = RGB(1)) = BoundaryFill(x, y, fill_value, boundary_value)
+BoundaryFill(x::Int = 1, y::Int = 1; fill_value::Colorant = RGB(1), boundary_value::Colorant = RGB(1)) = BoundaryFill(x, y, fill_value, boundary_value)
+BoundaryFill(p::CartesianIndex{2}; fill_value::Colorant = RGB(1), boundary_value::Colorant = RGB(1)) = BoundaryFill(p[1], p[2], fill_value, boundary_value)
+BoundaryFill(p::Point; fill_value::Colorant = RGB(1), boundary_value::Colorant = RGB(1)) = BoundaryFill(p.y, p.x, fill_value, boundary_value)
```
johnnychen94, line 228: > The pixels to define the boundary using verts is filled with this colorant parameter. 
Without this parameter, the whole picture will be filled with the `fill_value`

The `boundary_value` is still `RGB(1)` if you pass `BoundaryFill(p; fill_value=Gray(1))`.

```suggestion
BoundaryFill(x::Int = 1, y::Int = 1; fill_value::Colorant = RGB(1), boundary_value::Colorant = fill_value) = BoundaryFill(x, y, fill_value, boundary_value)
BoundaryFill(p::CartesianIndex{2}; fill_value::Colorant = RGB(1), boundary_value::Colorant = fill_value) = BoundaryFill(p[1], p[2], fill_value, boundary_value)
BoundaryFill(p::Point; fill_value::Colorant = RGB(1), boundary_value::Colorant = fill_value) = BoundaryFill(p.y, p.x, fill_value, boundary_value)
```

Need to add a test to check this.

src/core.jl
```diff
@@ -207,9 +206,10 @@ using ImageDraw
 img = zeros(RGB, 7, 7)
 expected = copy(img)
 expected[2:6, 2:6] .= RGB{N0f8}(1)
+
 verts = [CartesianIndex(2, 2), CartesianIndex(2, 6), CartesianIndex(6, 6), CartesianIndex(6, 2), CartesianIndex(2,2)]
 
-draw(img, verts, BoundaryFill(x = 4, y = 4, fill_value = RGB(1), boundary_value = RGB(1)); closed = true)
+draw(img, verts, BoundaryFill(4, 4; fill_value = RGB(1), boundary_value = RGB(1)); closed = true)
```
johnnychen94, line 212:
```suggestion
fill_method = BoundaryFill(4, 4; fill_value = RGB(1), boundary_value = RGB(1))
draw(img, verts, fill_method; closed = true)
```

johnnychen94 (commented):
If I don't miss anything from previous comments, this should be the last round of review. Once it's fixed/answered, it should be good to merge.

You're starting to write good tests, that's good news :tada: Do keep in mind: all reliable functionality is supported by more lines of codes of test cases.

src/polygonfill2d.jl
```diff
@@ -43,23 +41,32 @@ using ImageDraw
...
     f::AbstractPolyFillAlgorithm;
     closed::Bool = false,
-    ) where {T<:Colorant}
+    ) where N
+
+    # Boundscheck for verts
+    @boundscheck a = map(x -> (checkbounds(Bool, img, x[1], x[2]) ? nothing : throw(DomainError(x, "Vertice $x outside the image array domain"))), verts)
+
+    # Boundscheck for seed point
+    @boundscheck checkbounds(Bool, img, f.x, f.y) ? nothing : throw(DomainError((f.x, f.y),"Seed ($(f.x), $(f.y)) outside the image array domain"));
```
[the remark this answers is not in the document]
ashwani-rathee: @johnnychen94 Actually domain error is thrown to properly stop the execution in current case but when I use the suggestions it does return me false for invalid points but it doesn't stop implementation.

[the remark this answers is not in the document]
johnnychen94: Oh sorry, no `Bool` here.

```julia
julia> x = rand(4, 4);

julia> checkbounds(x, 5, 1)
ERROR: BoundsError: attempt to access 4×4 Matrix{Float64} at index [5, 1]
Stacktrace:
 [1] throw_boundserror(A::Matrix{Float64}, I::Tuple{Int64, Int64})
   @ Base ./abstractarray.jl:651
 [2] checkbounds(::Matrix{Float64}, ::Int64, ::Int64)
   @ Base ./abstractarray.jl:616
 [3] top-level scope
   @ REPL[2]:1

julia> checkbounds(Bool, x, 5, 1)
false
```

src/polygonfill2d.jl
```diff
@@ -56,11 +56,10 @@ function draw!(
     ) where N
 
     # Boundscheck for verts
-    @boundscheck a = map(x -> (checkbounds(Bool, img, x[1], x[2]) ? nothing : throw(DomainError(x, "Vertice $x outside the image array domain"))), verts)
-
+    @boundscheck map(x -> (checkbounds(Bool, img, x[1], x[2]) ? nothing : throw(DomainError(x, "Vertice $x outside the image array domain"))), verts)
```
johnnychen94, line 59: Oh boy, you can also change this to

```suggestion
@boundscheck foreach(x->checkbounds(img, x[1], x[2]), verts)
```

johnnychen94 (approved):

src/polygonfill2d.jl
ashwani-rathee: Yeah, I could have done that but that doesn't give information(in a concise way) to user what to correct 😕 
Current expected error looks far more informative in my understanding I think.
```julia
    Expected: DomainError(CartesianIndex(0, 6), "Vertice CartesianIndex(0, 6) outside the image array domain")
      Thrown: BoundsError(ColorTypes.RGB[RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0); RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0); RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0); RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0); RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0); RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0); RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0) RGB{N0f8}(0.0,0.0,0.0)], (CartesianIndex(0, 6),))
Stacktrace:

```

PR #60, opened 2021-05-15
WIP: Flood Fill Implementation

ashwanirathee:
Initial Base

Tasks:
- [ ] Lots of tests
- [x] Buggy implementation fix(it works for some cases, in some others it doesn't) 
- [x] Doc string of flood fill
- [x] Documentation of both boundary fill and flood fill


ashwani-rathee:
When I don't specify a kwarg in the algorithm( which will be really optional for FloodFill as it won't really need it), Ambiguity error is thrown. The draw in core.jl is very general, so there is that.
```julia
julia> res = draw(img, verts, FloodFill(350,200,fill_value =RGB(0), current_value = img[200,350]))
ERROR: MethodError: draw(::Matrix{RGB{FixedPointNumbers.N0f8}}, ::Vector{CartesianIndex{2}}, ::FloodFill{RGB{FixedPointNumbers.N0f8}}) is ambiguous. Candidates:
  draw(img::AbstractMatrix{T}, args...) where T<:Colorant in ImageDraw at /home/ashwani/julia-related/ImageDraw.jl/src/core.jl:385
  draw(img::AbstractArray{var"#s36", N} where var"#s36"<:Colorant, verts::Array{CartesianIndex{N}, 1}, f::ImageDraw.AbstractPolyFillAlgorithm; kwargs...) where N in ImageDraw at /home/ashwani/julia-related/ImageDraw.jl/src/polygonfill2d.jl:118
Possible fix, define
  draw(::AbstractMatrix{T}, ::Vector{CartesianIndex{2}}, ::ImageDraw.AbstractPolyFillAlgorithm) where T<:Colorant
Stacktrace:

```

src/polygonfill2d.jl
```diff
@@ -22,6 +22,26 @@ function (f::BoundaryFill)(
...
+
+"""
+
+function (f::FloodFill)(
+    res::AbstractArray{T,2},
+    verts::Vector{CartesianIndex{2}},
+    x::Int,
+    y::Int,
+    ) where {T<:Colorant}
+    if (res[y, x] != f.current_value || res[y, x] == f.fill_value) return end
```
johnnychen94, line 36: Can we get `current_value` through something like `res[x, y]` (or `res[y,x]`)?

johnnychen94 (commented):
> When I don't specify a kwarg in the algorithm( which will be really optional for FloodFill as it won't really need it unless we make it to be so), Ambiguity error is thrown. 

It looks like the ambiguity is triggered because the type annotation of `img` is not the same; so Julia can't figure out an order here.

Does this patch fix it?

```diff
- draw(img::AbstractArray{T,2}, args...) where {T<:Colorant} = draw!(copy(img), args...)
+ draw(img::AbstractArray{T}, args...; kwargs...) where {T<:Colorant} = draw!(copy(img), args...; kwargs...)
```

src/polygonfill2d.jl
ashwani-rathee: Yes, that can be done. We want to set the value of current_value at time when struct creation is called, but at that time res is not accessible  to struct . There will be quick way to solve it 😄 , I need to lookup .

src/polygonfill2d.jl
```diff
@@ -28,12 +28,12 @@ end
...
+    verts::Vector{CartesianIndex{N}},
     x::Int,
     y::Int,
-    ) where {T<:Colorant}
-    if (res[y, x] != f.current_value || res[y, x] == f.fill_value) return end
+    ) where {T<:Colorant, N}
+    if (res[y, x] != f.current_value || res[y, x] == f.fill_value) return res end
     if checkbounds(Bool, res, y, x) res[y, x] = f.fill_value end
     f(res, verts, x + 1, y)
     f(res, verts, x - 1, y)
```
johnnychen94, line 39: This can still be `{T, 2}` as the implementation is specified for matrix; this won't affect the method ambiguity on `draw` so no need to change.

Issue #63, opened 2021-05-29
Add logo for document


hyrodium:
Hi! i'm now planning to add logo images to ImageDraw.jl.

![logo](https://user-images.githubusercontent.com/7488140/120065676-7af9d280-c0ad-11eb-93a0-d3f57e16ea4a.png)
![logo-dark](https://user-images.githubusercontent.com/7488140/120065675-7a613c00-c0ad-11eb-87c9-6d8d256af185.png)

![image](https://user-images.githubusercontent.com/7488140/120066902-be573f80-c0b3-11eb-94f7-b7d538d8d013.png)
![image](https://user-images.githubusercontent.com/7488140/120066910-c7481100-c0b3-11eb-9636-4ba1f9d1a8b0.png)

```julia
using Images, ImageDraw

# original image
img = load(download("https://raw.githubusercontent.com/JuliaImages/juliaimages.github.io/source/docs/src/assets/logo.png"))

# Define some colors
c_r, c_g, c_b, c_p = RGBA.([Colors.JULIA_LOGO_COLORS...])
gray_dark = RGBA{N0f8}(0.1,0.1,0.1)
gray_light = RGBA{N0f8}(0.9,0.9,0.9)

# change base corlor black to gray
img = img*N0f8(0.5) .+ RGBA{N0f8}(0.5,0.5,0.5,0)
img_dark = copy(img)
img_light = copy(img)

# define points on the image
p_green = Point(87,35)
p_red = Point(75,55)
p_purple = Point(99,55)
p1 = Point(13,96)
p2 = Point(52,50)
p3 = Point(73,73)
p4 = Point(84,62)
p5 = Point(116,97)

for (img,gray,name) in ((img_dark,gray_light,"logo-dark.png"), (img_light,gray_dark,"logo.png"))
    draw!(img, LineTwoPoints(p1,p2), gray)
    draw!(img, LineTwoPoints(p2,p3), gray)
    draw!(img, LineTwoPoints(p3,p4), gray)
    draw!(img, LineTwoPoints(p4,p5), gray)

    draw!(img, Cross(p1,8), gray)
    draw!(img, Cross(p2,8), gray)
    draw!(img, Cross(p3,8), gray)
    draw!(img, Cross(p4,8), gray)
    draw!(img, Cross(p5,8), gray)

    draw!(img, Ellipse(CirclePointRadius(p_green,10)), c_g)
[...]

PR #64, opened 2021-05-30
Add logo to document

hyrodium:
See more info at #63.


docs/logo.jl
```diff
@@ -0,0 +1,47 @@
...
+    p_purple = Point(99,55)
+    p_1 = Point(13,96)
+    p_2 = Point(52,50)
+    p_3 = Point(73,73)
+    p_4 = Point(84,62)
+    p_5 = Point(116,97)
+
+    # Define path
+    path_dark = "docs/src/assets/logo-dark.png"
+    path_light = "docs/src/assets/logo.png"
```
johnnychen94, line 27: The `docs/make.jl` file might be called from two typical paths:

- `project_dir`: `include("docs/make.jl")`
- `project_dir/docs`: `include("make.jl")`

The current version only handles the first case. To also support the second case, might be good to use `@__DIR__` macro here, which will be automatically expanded to `@__DIR__() = "project_dir/docs"`

Also use `joinpath` for windows users.

```suggestion
    path_dark = joinpath(@__DIR__, "src", "assets", "logo-dark.png")
    path_light = joinpath(@__DIR__, "src", "assets", "logo.png")
```

docs/logo.jl
```diff
@@ -0,0 +1,47 @@
+function generate_logo()
+    # Load original image (from https://github.com/JuliaImages/juliaimages.github.io/blob/source/docs/src/assets/logo.png)
+    img = load("docs/logo_original.png")
```
johnnychen94, line 3:
```suggestion
    img = load(joinpath(@__DIR__, "logo_original.png"))
````

johnnychen94 (approved):
Once the path issue get fixed, I think it's good to merge

docs/logo.jl
hyrodium: Thanks for the suggestion!

PR #70, opened 2022-07-12
Filled Polygon Option 

opiateblush:
- added simple fill option for polygons, similar to the filled ellipse
- the polygon's boundary must be drawn even in the filled case, as the scanline filling algorithm cannot handle multiple vertices on a straight line

### DIFF
--- a/src/paths.jl
+++ b/src/paths.jl
@@ -13,13 +13,50 @@ end
 
 #Polygon methods
 
-Polygon(v::AbstractVector{Tuple{Int, Int}}) = Polygon([Point(p...) for p in v])
-Polygon(v::AbstractVector{CartesianIndex{2}}) = Polygon([Point(p) for p in v])
-Polygon(rectangle::RectanglePoints) = Polygon([rectangle.p1, Point(rectangle.p2.x, rectangle.p1.y), rectangle.p2, Point(rectangle.p1.x, rectangle.p2.y)])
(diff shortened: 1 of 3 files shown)


opiateblush:
this might partly solve #53

johnnychen94 (approved):
It looks good to me (the intuitive idea), but I'd like to have @ashwani-rathee double-check before merging it because this PR overlaps with #60

opiateblush:
It's been quite some time, is there any problem with my implementation? I'm fairly new to the pull request business, so I hope it's ok to ask if there is something wrong.

babaq:
This PR would be very useful for us to fill contours, instead of PyCalling to skimage.draw polygon of our current approach.

babaq:
Can we merge this branch and tag a new release, seems straightforward to review the small amount of code change.

src/paths.jl
```diff
@@ -13,13 +13,54 @@ end
...
+
+    if polygon.fill
+        polygonscanfill!(img, polygon, color)
+    end
+
+    img
+end
+
+function polygonscanfill!(img::AbstractArray{T, 2}, polygon::Polygon, color::T) where T <: Colorant
+    image_height, image_width = size(img)
```
timholy, line 32: Could we use `axes` here and check both the minimum and maximum indices? Not all AbstractArrays start indexing at 1 (e.g., OffsetArrays.jl) The `clamp` arguments below would also have to change, of course.

timholy (commented):
Looks good, thanks!

opiateblush:
Those commits were from me, btw. Unfortunately, i made them with a different account, which I usually use for work stuff.

PR #75, opened 2023-07-23
ImageDraw 0.2.6

IanButterworth:
@timholy I'm basing this just being a patch release on https://github.com/JuliaImages/Images.jl/issues/1036#issue-1816855105

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ImageDraw"
 uuid = "4381153b-2b60-58ae-a1ba-fd683676385f"
-version = "0.2.5"
+version = "0.2.6"
 
 [deps]
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"


timholy:
#58 seems likely to be breaking (in a very good way), so I think you'll need a minor bump.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ImageDraw"
 uuid = "4381153b-2b60-58ae-a1ba-fd683676385f"
-version = "0.2.5"
+version = "0.2.6"
```
IanButterworth, line 3:
```suggestion
version = "0.3.0”
```

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ImageDraw"
 uuid = "4381153b-2b60-58ae-a1ba-fd683676385f"
-version = "0.2.5"
+version = "0.3.0”
```
IanButterworth, line 3:
```suggestion
version = "0.3.0"
```

timholy (approved):

IanButterworth:
@timholy I'm only now noticing that https://github.com/JuliaImages/ImageDraw.jl/pull/58 was released in v0.2.5 
Should we still keep this as 0.3.0 (i.e. fix what should've been a breaking release 2+ years ago?) or revert this and release v0.2.6

timholy:
Oh, then non-breaking is fine. Sorry about that, I don't know what I was looking at!

PR #77, opened 2023-09-25
Semi-transparent shapes and automatic flood fill.

norpadon:
### Add an option to make drawn shapes semi-transparent by specifying a keyword argument `opacity`, like this:
```julia
ImageDraw.draw!(image, ImageDraw.Cross(ImageDraw.Point(x, y), 3), color, opacity=0.2)
```

### Also change the polygon filling API. Now you don't need to pass any objects containing seed coordinates, just specify keyword arguments `fill_color` and `fill_opacity`:
```julia
ImageDraw.draw(keypoint_vis, rectangle, RGBA{Float64}(1, 0, 0), opacity=1.0, fill_color=RGBA{Float64}(0, 1, 0), fill_opacity=0.5)
```

keypoint_vis


johnnychen94:
The new feature looks pretty nice!

This PR introduces a few breaking things, I insist you trying both of the following before merging:

- use `@deprecate` and `Base.depwarn` to properly handle methods/types that could still be supported in a compatible way. For deprecation, [This](https://github.com/JuliaImages/ImageTransformations.jl/pull/138) is a good reference if you've never done such kind of thing before.
- introduce a good changelog.md file to let future users what's breaking, the rational, and the suggested steps to migrate. [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl/blob/master/CHANGELOG.md) does this really really well.

Sixzero:
I would like to know why is it good to introduce the flag opacity, instead of using RGBA{N0f8}(1.,0,0,0.4) for 40% opacity? I mean if the color already defines the opacity, then why define it in the function call?

PR #79, opened 2023-12-08
fix #78

### LINKED ISSUE #78: Why collect the indices instead of drawing directly?
In
https://github.com/JuliaImages/ImageDraw.jl/blob/8b2a4f125cdaf4031b2e72755910ee39b094fccc/src/ellipse2d.jl#L18
we collect indices to later iterate over and draw ([here](https://github.com/JuliaImages/ImageDraw.jl/blob/8b2a4f125cdaf4031b2e72755910ee39b094fccc/src/ellipse2d.jl#L23)). Why not skip the collection and just draw right there and then inside the loop? 

It would avoid dynamically allocating `x` and `y`.

If there's no reason not I can PR and benchmark.

yakir12:
This PR drammatically improves on performance for ellipse drawing:

```julia
using Revise
using BenchmarkTools
using ImageDraw 
import ColorTypes: RGB, N0f8

img = rand(RGB{N0f8}, 1000, 1000)
c = CirclePointRadius(Point(500, 500), 400)
f() = draw!(img, c)
@benchmark f()
```
before this PR:
```julia
BenchmarkTools.Trial: 1875 samples with 1 evaluation.
 Range (min … max):  2.258 ms …   5.262 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.499 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.600 ms ± 292.306 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▅██▆▂▁                                                  
  ▂▃▄▇▇███████▆▇▅▆▇██▆▅▄▃▃▃▃▃▃▃▂▂▁▂▂▂▂▂▃▃▃▃▃▃▃▂▂▂▂▂▃▃▂▂▂▂▁▁▂▂ ▃
  2.26 ms         Histogram: frequency by time        3.63 ms <

 Memory estimate: 3.66 MiB, allocs estimate: 22.
```
After this PR:
```julia
BenchmarkTools.Trial: 8135 samples with 1 evaluation.
 Range (min … max):  594.081 μs …  1.619 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     609.660 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   612.257 μs ± 26.233 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

               █        ▁                                       
  ▂▂▂▂▁▁▂▂▃▃▂▃▃█▄▄▄▇▆▅▅▄█▄▄▄▄▄▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  594 μs          Histogram: frequency by time          643 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
````

### DIFF
--- a/src/ellipse2d.jl
+++ b/src/ellipse2d.jl
@@ -5,8 +5,6 @@ Ellipse(p::CartesianIndex{2}, args...; kwargs...) = Ellipse(Point(p), args...; k
 Ellipse(circle::CirclePointRadius) = Ellipse(circle.center, circle.ρ, circle.ρ; thickness = circle.thickness, fill = circle.fill)
 
 function draw!(img::AbstractArray{T, 2}, ellipse::Ellipse, color::T) where T<:Colorant
-	ys = Int[]
-	xs = Int[]
 	break_point = 0
 	if ellipse.fill == false
 		break_point = ((ellipse.ρy - ellipse.thickness) / ellipse.ρy) ^ 2 + ((ellipse.ρx - ellipse.thickness) / ellipse.ρx) ^ 2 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by yakir12** [comment]
Does anyone have time to review this? The code change is pretty small and the improvement is substantial, to put it in relative terms :stuck_out_tongue:

File: src/ellipse2d.jl [hunk 1]
```diff
@@ -15,16 +13,14 @@ function draw!(img::AbstractArray{T, 2}, ellipse::Ellipse, color::T) where T<:Co
 		for j in ellipse.center.x : ellipse.center.x + ellipse.ρx
 			val = ((i - ellipse.center.y) / ellipse.ρy) ^ 2 + ((j - ellipse.center.x) / ellipse.ρx) ^ 2
 			if val < 1 && val >= break_point			
-				push!(ys, i)
-				push!(xs, j)
+        yi = Int(i)
```
ashwanirathee: Hey yakir, please format it correctly, otherwise it looks good to me

mkitti:
Looks good to me. The indenting  looks a bit strange from my phone though.

src/ellipse2d.jl
yakir12: oh wow, I was really confused cause it looked fine in vim but off when I printed it out. Apparently some of the spaces here are tabs, which I think is discouraged (precisely for this reason).

yakir12:
This file contained tab-spaces so I replaced them with (2) spaces, hope that doesn't go against the style guide of ImageDraw...

src/ellipse2d.jl
mkitti: Thanks... but I think you used two spaces. The rest of the package uses four.

yakir12: Right, fixed it now.

